PR #3, opened 2018-03-07
Feature/use new swagger api

domharrington:
Updated the readme too https://github.com/readmeio/rdme/blob/2ff1828c696d762d9e8a7c35d493698ff56f49e7/README.md


lib/swagger.js
```diff
@@ -0,0 +1,49 @@
...
+var fs = require('fs');
+var path = require('path');
+
+exports.swagger = true;
+exports.login = true;
+exports.desc = 'Upload your swagger file to ReadMe';
+exports.category = "services";
+
+exports.run = function(config, info, cb) {
+  if (!cb) cb = (err) => {
```
gkoberger: Probably fine, but we should watch for people complaining that they're on an old vrsion of Node and this doesn't work.

domharrington: 👍 just looked this up, arrow functions have been in node since 4, which is coming to it's end of life for support at the end of april

PR #4, opened 2018-05-07
Feature/docs upload

domharrington:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright © 2016 ReadMe.io
+Copyright © 2018 ReadMe.io
 Copyright © 2011-2012 Paul Vorbach
```
mjcuva: Who is Paul?

gkoberger: We can remove this! He's had nothing to do with this module at all.

PR #35, opened 2019-06-18
Next steps after you import a swagger file in the cli

gratcliff:
1. Updated endpoints to reflect API deprecation of Swagger handlers in main Readme app.
2. Minor logic update to success function to pull relevant data and present updated log string.
3. Updated tests to reflect endpoint change.

PR #36, opened 2019-06-22
If an OAS is not supplied, attempt to discover one in the root dir.

erunion:
This makes the OAS argument of `rdme swagger` optional. If one is not supplied, we will now attempt to locate a `(swagger|openapi).(json|yaml)` file in the directory that `rdme swagger` is being run within. If one is then not found, we'll now point the user in the direction of using `rdme oas init` to get started down the path to spec creation.

I've also fixed a minor bug where if you supplied an incorrect API key to `rdme swagger`, we will now fail with a clearer error message than "There was an error uploading!".

### DIFF
--- a/lib/swagger.js
+++ b/lib/swagger.js
@@ -21,37 +21,64 @@ exports.run = function({ args, opts }) {
     return Promise.reject(new Error('No api key provided. Please use --key'));
   }
 
-  if (!args[0]) {
-    return Promise.reject(
-      new Error('No swagger file provided. Usage `rdme swagger <swagger-file>`'),
-    );
-  }
-
-  function success(data) {
-    console.log(data);
-    console.log('Success! '.green);
-  }
-
-  function error(err) {
-    try {
-      return Promise.reject(new Error(JSON.parse(err.error).description));
-    } catch (e) {
-      return Promise.reject(new Error('There was an error uploading!'));
+  function callApi(specPath) {
+    function success(data) {
+      console.log(data);
+      console.log('Success! '.green);
     }
+
+    function error(err) {
+      try {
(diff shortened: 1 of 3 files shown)


lib/swagger.js
```diff
@@ -21,37 +21,64 @@ exports.run = function({ args, opts }) {
...
+        return;
+      }
+
+      console.log(`We found ${file} and are attempting to upload it.`.yellow);
+
+      resolve(callApi(file));
+    });
+
+    reject(
+      new Error(
```
mjcuva: I feel like if people are running this command, they likely have a swagger file, maybe it would be better to tell them how to provide a path as well? Something like “We couldn’t find a swagger or OpenAPI file. Run `rdme swagger ./path/to/file` to upload an existing file or `rdme oas init` to create a fresh one!”

erunion: Good idea.

lib/swagger.js
```diff
@@ -21,37 +21,64 @@ exports.run = function({ args, opts }) {
...
 
-  function success(data) {
-    console.log(data);
-    console.log('Success! '.green);
-  }
+    function error(err) {
+      try {
+        if (err.statusCode === 401) {
+          return Promise.reject(
+            new Error('There was a problem with your api key. Please try again.'),
```
domharrington: I've actually added this in the readme http api, you can see it over in #1553 over there! So we probably don't need this anymore.

erunion: I'll pull it here and we can just rely on that change when that gets deployed.

erunion:
@domharrington @gratcliff Merged your changes in from https://github.com/readmeio/rdme/pull/35 and this should all good to go now.

PR #44, opened 2019-07-16
cli error handling

gratcliff:
1. Util folder added
2. Error handling bundled
3. CLI args modified


lib/versions/create.js
```diff
@@ -12,15 +13,13 @@ exports.run = async function({ opts }) {
...
-  if (!key) {
-    return Promise.reject(new Error('No api key provided. Please use --key'));
-  }
-
-  if (!version) {
-    return Promise.reject(
-      new Error('No version provided. Please specify a semantic version using --version'),
-    );
-  }
+  await validateErrors([
```
dok: I think it's a step in the right direction! I think there was some misunderstanding, probably because of the way I explained my thoughts in the other CR.

I was mostly worried about code-reuse. I do see several duplicate lines for the same error message in the versions `create|delete|update` files. So I think it would be good to place them inside a single location. In this file currently, it seems the messages are still being duplicated across multiple files.

gratcliff: Yup, and the need to be the way that the code is run. `require('./cli')(parseArgs._[0] || 'help', parseArgs._.slice(1), parseArgs)` References individual files. Either we can migrate the whole project to class based, or we can use a helper function like the one created in this PR.

dok: We can leverage the object `opts` here. What about something like:

```
validateErrors(opts, 'key', 'version')
```

and have the message stored in a single place? (potentially errorhandler.js)
```
exports.validators = {
  key: [opts => opts.key, 'No api key provided. Please use --key'],
}
```

Just a thought!

gratcliff: I think there's a couple considerations that need to take place:
a) a level of customization that may need to take place on a message by message basis
b) option flags are not always a requirement

dok: I agree with you, it's going to need a systematic change. Can we hold off on merging this implementation? We can take a look at another time when it's fitting to put in some validation frameworking within `cli.js` We'll be able to refer back to this

gratcliff: Works for me. I'll cancel this PR and leave the branch up.

PR #49, opened 2019-07-20
🌺 Quality of life improvements

erunion:
rdme as it exists has a horde of usability issues:

* No commands are documented.
* There's no way to get help for individual commands (and they aren't documented anyway).
* Error handling is all over the place. Sometimes we return nice, colored, errors, others we dump raw JSON responses to the console.

There's **got** to be a better way!

1CDX

### Improvements
* [x] All commands have been fully documented.
* [x] More command categories/groups have been created: "Administration", "Sync Swagger/OpenAPI files", "Documentation", and "Versions"
* [x] Heavily improved error handling to account for the wide variety of error cases that might happen (API errors, type errors, bad command argument errors, unexpected errors, etc.)
* [x] Cleaned up a lot of the errors that we were returning so they're more consistently represented.
* [x] 🆕 Added support for getting help via `rdme <command> --help` and `rdme help <command>`
* [x] 🆕 Related commands are now surfaced on command help screens. So for example if you do `rdme help versions`, you'll see a related commands list of every other versions-related command available.
* [x] 🆕 Owlbert has been added to the main `rdme help` screen because everyone needs more Owlbert.

### In action
```
$ ./rdme.js

          ░                                        ░
          ░░                                       ░░
        ░░░░░                                     ░░░░ ░
       ░░░░░░░                                    ░░░░░░
       ▐░░░░░░▄                                 ░░░░░░░░░
       ░░░░░░░░░▐░░░                        ░░░░░░░░░░░░░   rdme
        ░░░░░░░░░░░░░░░░░▐▐▐░░░░░░░░░░▐▐░░░░░░░░░░░░░░░░░░
      ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░    a utlity for interacting with ReadMe.io
      ▐░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
      ▐░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
      ▐░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▄▄▄▄░░░
      ▐░░░░░░░░░░░░░░░▄█▓▓▓▓▓▓█▄░░░░░░░░░░░░░░▄█▓▓▓▓▓▓▓▓▓▄░          ░░
[...]

PR #53, opened 2019-08-03
Adding new authentication oriented commands

erunion:
This adds two new commands: 

- `rdme logout` - Logs the current user out and clears their saved config file.
- `rdme whoami` - Returns the email and project of the user that's currently logged in.

I also did a few minor maintenancey things to clean some things up from https://github.com/readmeio/rdme/pull/49:

- Renaming the `weight` property in commands to `position`. Since our new `help` screen shows the split groups, the weight is really more of a position declaration in that screen now so I've renamed it to reflect that. It still works the same, this is just purely renaming the property so it's clearer what it represents.
- Making our test descriptions `describe('...' () => {})` consistent.

### DIFF
--- a/cmds/logout.js
+++ b/cmds/logout.js
@@ -2,22 +2,22 @@ const config = require('config');
 const configStore = require('../lib/configstore');
 const loginCmd = require('./login');
 
-exports.command = 'whoami';
-exports.usage = 'whoami';
-exports.description = 'Displays the current user and project authenticated with ReadMe.';
+exports.command = 'logout';
+exports.usage = 'logout';
+exports.description = 'Logs the currently authenticated user out of ReadMe.';
 exports.category = 'admin';
-exports.weight = 3;
+exports.weight = 2;
 
 exports.args = [];
 
-exports.run = () => {
+exports.run = async () => {
   if (!configStore.has('email') || !configStore.has('project')) {
     return Promise.reject(new Error(`Please login using \`${config.cli} ${loginCmd.command}\`.`));
   }
 
+  configStore.clear();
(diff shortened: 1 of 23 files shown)


cmds/logout.js
```diff
@@ -0,0 +1,23 @@
...
+exports.usage = 'logout';
+exports.description = 'Logs the currently authenticated user out of ReadMe.';
+exports.category = 'admin';
+exports.position = 2;
+
+exports.args = [];
+
+exports.run = async () => {
+  if (!configStore.has('email') || !configStore.has('project')) {
+    return Promise.reject(new Error(`Please login using \`${config.cli} ${loginCmd.command}\`.`));
```
gratcliff: This error message doesn't make a whole lot of sense. Why would we prompt a user to login so that they can log out?

test/cli.test.js
```diff
@@ -88,7 +88,7 @@ describe('cli', () => {
     });
 
     it('should not show related commands on commands that have none', () => {
-      cli(['login', '--help']).then(output => {
+      cli(['swagger', '--help']).then(output => {
```
gratcliff: Is there a reason for this change?

erunion: `login` now has related commands, where as `swagger` has none.

cmds/logout.js
erunion: Hm yeah, I guess we can just show the "you have logged out" message if they aren't logged in.

test/cli.test.js
gratcliff: Makes sense.

PR #54, opened 2019-08-06
Feature: PDF generation

wishpinebetter:
Now it generates `/docs` and `/references` pages into a single PDF file.


brandon1980:
@bsc1780

Issue #55, opened 2019-08-07
How to set boolean flags to false for automation


lpww:
Hey! 

The [docs on automation](https://www.npmjs.com/package/rdme#automating-this-process) state that you should pass values to prevent being prompted for manual intervention during version creation. The example only shows setting the booleans to true. How do you set them to false?

I've tried `--main=0`, `--main=false`, `--main=False`, and `-n` but all give errors eg `Unknown value: --main=0`. I'm not sure if this is a bug or just a missing piece of documentation?

gratcliff:
Hey there! Thanks for checking out our CLI tool. 

Typically, you should be able to pass `false` as a flag value in the same way that you can pass `true`, ala `--main=false`. I haven't been able to replicate the bug you're encountering. Using false is working as intended. 

<img width="1628" alt="Screen Shot 2019-08-07 at 10 54 31 AM" src="https://user-images.githubusercontent.com/8877659/62646051-52b24a00-b902-11e9-9c67-89435f2aa5d6.png">

Let's see if we can get to the bottom of this, though! Can you provide me information on your node.js, npm , and rdme versions, along with the exact command (sans api key) you used to produce this?

Thanks very much!

lpww:
Hi @gratcliff, 

I think I see the issue. The example in the doc passes the `version` with no flag. For some reason this doesn't work when setting flags to false. If I do `--version=v0.3.1`, as you have in your example, I can set other flags to false successfully.

So, this fails:
```
rdme versions:create v0.3.1 --fork=v0.2.0 --codename=thomastest --isPublic --main=false
Unknown value: --main=false
```

This works:
```
rdme versions:create --version=v0.3.1 --fork=v0.2.0 --codename=thomastest --isPublic --main=false --beta=false
Version v0.3.1 created successfully.
```

I think this is a bug but you have given me a workaround - thanks!

lpww:
Hey @gratcliff,

I just realized that the workaround is actually not working. It doesn't give an error but when checking the ui, I can see that all the values are `true` instead of `false`. There are the versions you asked for:

```
node -v
v10.16.0

npm -v
6.9.0

rdme -v
3.4.0
```

This is the failing command:

```
rdme versions:create --version=v0.3.1 --fork=v0.2.0 --codename=thomastest --isPublic=false --main=false --beta=false
```

Here is a screenshot of the ui after creating the new version:

[...]

PR #57, opened 2019-08-08
fix/version-flags

gratcliff:
Fix in response to https://github.com/readmeio/rdme/issues/55

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ If you wish to see the raw output from our API in this response, supply  the `--
 
 #### Create a new version
 ```sh
-rdme versions:create <version>
+rdme versions:create --version={project-version}
 ```
 
 ##### Automating this process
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -105,7 +105,7 @@ If you wish to see the raw output from our API in this response, supply  the `--
 
 #### Create a new version
 ```sh
-rdme versions:create <version>
+rdme versions:create --version={project-version}
```
erunion: Technically both of these are correct since `version` is configured as a `defaultOption` on these commands.

gratcliff: Checkout the issue I linked, they were not being parsed in the same manner due to how the cli package handles booleans interacting with multiflag. Since they've been shifted to strings, it seems to be working as you mentioned. I'll reflect this :)

PR #58, opened 2019-08-08
Add x-readme-source header

mjcuva:
Add header to tell ReadMe that the synced file is coming from the cli

### DIFF
--- a/cmds/swagger.js
+++ b/cmds/swagger.js
@@ -92,6 +92,7 @@ exports.run = async function(opts) {
       },
       headers: {
         'x-readme-version': versionCleaned,
+        'x-readme-source': 'cli',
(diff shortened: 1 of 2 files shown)


erunion:
@mjcuva This LGTM. Do we need to wait for the `cli` header support in the app to get into prod before releasing this?

domharrington:
I don't think it'll have any adverse affects if this gets released before that gets merged. Right now it just defaults to `cli`.

gratcliff:
@domharrington is correct, it's hardcoded in `dash/api/v1/api-specification` right now at the base `POST` request.

erunion:
Confirmed! Merging and tagging a new release.

Issue #69, opened 2019-09-06
Remove installation warnings


michaelfarrell76:
I get the following warnings when I install with yarn, the only warnings in our entire monorepo are being emitted by readmeio packages

```
warning rdme > opn@6.0.0: The package has been renamed to `open`
```

erunion:
Resolved in 3.4.5

michaelfarrell76:
I'm getting new warnings when upgrading to 3.4.5

```
warning rdme > oas > base-store > data-store > cache-base > set-value@0.4.3: Critical bug fixed in v3.0.1, please upgrade to the latest version.
warning rdme > oas > base-store > data-store > union-value > set-value@0.4.3: Critical bug fixed in v3.0.1, please upgrade to the latest version.
```

erunion:
@michaelfarrell76 Whoops, I forgot to update our `oas` dependency after fixing that issue there. This should all be resolved in 3.4.6 now.

PR #72, opened 2019-09-09
Replacing the opn module with open since it was renamed.

erunion:
Resolves https://github.com/readmeio/rdme/issues/69


gratcliff:
Was this just renamed? I installed in somewhere else last week under the name `opn`

erunion:
Yeah looks like it was renamed back in March. https://github.com/sindresorhus/open/releases/tag/v6.0.0

PR #78, opened 2019-09-17
fix/help-semver-validation

gratcliff:
Ticket addresses:
1.  `help` flag for `versions:create`, `versions:delete`, and 'versions:update` by fixing parse error for dependency
2. Adding support for semver validation in prompt and directly when using the cli with flags.


erunion:
@gratcliff Tagged to 3.4.7

gratcliff:
> @gratcliff Tagged to 3.4.7

Thank you sir.

PR #218, opened 2020-07-15
feat: overhauling error handling to support the new api error responses

erunion:
## 🧰 What's being changed?

This overhauls every command to support the new way we're handling errors within the API and allowing us to surface Metrics log IDs in the console for API failures

### DIFF
--- a/cmds/docs/index.js
+++ b/cmds/docs/index.js
@@ -5,6 +5,7 @@ const config = require('config');
 const crypto = require('crypto');
 const frontMatter = require('gray-matter');
 const { promisify } = require('util');
+const APIError = require('../../lib/apiError');
 
 const readFile = promisify(fs.readFile);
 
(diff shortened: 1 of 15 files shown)


cmds/docs/index.js
```diff
@@ -49,6 +50,9 @@ exports.run = function (opts) {
   }
 
   const files = fs.readdirSync(folder).filter(file => file.endsWith('.md') || file.endsWith('.markdown'));
+  if (files.length === 0) {
+    return Promise.reject(new Error(`We were unable to locate Markdown files in ${folder}.`));
```
erunion, line 54: Adding this error message because if a user tries to use `rdme docs` with a directory that has subdirectories containing Markdown, we currently can't handle those Markdown files as `rdme docs` can't handle category creation and management right now.

https://app.asana.com/0/1178132843162889/1183431246035890/f

Issue #251, opened 2020-10-20
Unable to update docs using docs command


rayraffi:
Hi, I am trying to use the docs command to create or update our guides in an automated way in a CI pipeline, but instead of updating a doc that exists, the CLI keeps creating a new one. I have the slug for the doc defined in the front matter of the markdown file and the file name matches this slug. 

Looking at the source code for the command [here](https://github.com/readmeio/rdme/blob/master/cmds/docs/index.js) it looks like when the `updateDoc` function is called it gets passed undefined for `existingDoc` parameter, which should raise an error in the if-check in the beginning of the function, which makes it always create the doc instead.

jesseyowell:
Hi there! There is definitely a bug with sync'ing diffrent types of doc slugs. Here's the format that works without issue:
`---
title: "files"
excerpt: "these are my files"
slug: "files"
category: 6039419eee15f1005fb57099
---

This is my page on files!`

The occurs when you're adding in a hyphen tp the slug, so we'll need to fix that functionality!

PR #312, opened 2021-05-25
Lowercase filenames to match default slugs

ollyfg:
Currently, all slugs are cast to lowercase on Readme.com.
However when syncing, it left the name as uppercase, which **always**
resulted in a 404, and thus a new page being created.

This might fix https://github.com/readmeio/rdme/issues/251.


erunion:
@ollyfg  Can you rebase the latest `main` into this branch? We've moved some things around in the process of dropping support for Node 10.

ollyfg:
There you go! I don't think anything moved in this one.

erunion (approved):

erunion:
Thank you! We're currently working on a major release within the next couple weeks so this'll go out then.

PR #325, opened 2021-06-09
feat: dropping support for node 10, expanding coverage to node 14/16

erunion:
## 🧰 Changes

🚨  This'll be a breaking change and will require a bump to `rdme@4`!

* [x] Drops support for Node 10
* [x] Expands our `engines`, and CI testing, coverage to Node 14 and Node 16.
* [x] Starts using NPM 7 in CI, and updates the lockfile to the same.
* [x] Upgrades a handful of dev dependencies to their latest release: `eslint`, `prettier`, `nock`, `@readme/eslint-config`, `jest`

## 🧬 QA & Testing

* [ ] Do all unit tests pass?


package.json
```diff
@@ -1,7 +1,12 @@
 {
-  "author": "ReadMe <support@readme.io> (https://readme.com)",
   "name": "rdme",
+  "version": "3.8.2",
   "description": "ReadMe's API CLI",
+  "author": "ReadMe <support@readme.io> (https://readme.com)",
+  "engines": {
+    "node": "^12 || ^14 || ^16"
+  },
+  "main": "index.js",
```
kanadgupta, line 9: What's this line for? Just checked this and https://github.com/readmeio/rdme/pull/327 and it looks like we don't even have an `index.js` file

kanadgupta (commented):
One small q below otherwise LGTM!

package.json
erunion: ah my bad

rahulhegdee (approved):

PR #327, opened 2021-06-19
chore: directory spring cleaning

erunion:
## 🧰 Changes

Well it ain't spring but still a good time to clean up the tree while we're working on a major release.

* [x] Updating our `LICENSE` to 2021 and dropping the `.io` in ReadMe.
* [x] Moves the entire app into a `src/` directory structure.
* [x] Renames the `test/` directory to `__tests__` to match how our other open source libraries are structured.

## 🧬 QA & Testing

* [ ] Tests pass?


.editorconfig
```diff
@@ -1,12 +0,0 @@
-root = true
```
kanadgupta, line 1: RIP lol

kanadgupta: We could probably do this in ReadMe proper as well?

bin/rdme
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env node
```
kanadgupta, line 1: I'm assuming it's fine that this file doesn't have an extension since we have this line?

kanadgupta (approved):

rahulhegdee (approved):

bin/rdme
erunion: yep

.editorconfig
erunion: Yeah we probably should remove it there if it's still there. Prettier takes care of all this for us now.

PR #332, opened 2021-06-26
fix: validate files before uploading in swagger command

rahulhegdee:
## 🧰 Changes

- Added validation for files prior to creating/updating them in the project
- Added test to check validation

## 🧬 QA & Testing

- [ ] Make sure that invalid files throw an error
- [ ] Make sure that valid files still upload successfully


erunion (approved):
nice

PR #333, opened 2021-06-30
feat: rename swagger command to openapi and keep swagger as alias

rahulhegdee:
## 🧰 Changes

- renames swagger command to openapi
- keeps swagger command as an alias to the openapi command

## QA & Testing
- check that both openapi and swagger commands work the same


src/cmds/swagger.js
```diff
@@ -1,18 +1,11 @@
...
-const { prompt } = require('enquirer');
-const OASNormalize = require('oas-normalize');
-const promptOpts = require('../lib/prompts');
-const APIError = require('../lib/apiError');
+const openapi = require('./openapi');
 
 exports.command = 'swagger';
 exports.usage = 'swagger [file] [options]';
-exports.description = 'Upload, or sync, your Swagger/OpenAPI file to ReadMe.';
+exports.description = 'Alias for the openapi command.';
```
erunion, line 6:
```suggestion
exports.description = 'Alias for `rdme openapi`.';
```

__tests__/cmds/openapi.test.js
```diff
@@ -0,0 +1,226 @@
+const nock = require('nock');
```
erunion, line 1: Looks like this is the full content of `swagger.test.js` so since the command functionality is the same we should be able to rename that to this instead and then add a test to make sure that hitting `rdme swagger` loads the `rdme openapi` command.

erunion (changes requested):

__tests__/cmds/openapi.test.js
rahulhegdee: Okay perfect, I wasn't sure if we were trying to test to make sure they worked the same. I'll change that, thanks!

erunion (approved):
i just merged #332 in so you've got a conflict now, but once that's resolved I'll pull this in!

PR #341, opened 2021-07-21
fix: command position sorting within the help screen

erunion:
## 🧰 Changes

This fixes some improper `Array.sort()` logic within the help screen where the `position` properties that each command emits weren't being properly utilized.

## 🧬 QA & Testing

Before:

Screen Shot 2021-07-20 at 11 20 17 PM

After:

Screen Shot 2021-07-20 at 11 20 25 PM


rahulhegdee (approved):
Nice!

PR #342, opened 2021-07-27
feat: add $ref bundling to OpenAPI/Swagger command

rahulhegdee:
## 🧰 Changes

Adds $ref bundling to the OpenAPI command so that all the references we need are inserted into the file we're bundling to. 
Adds support for uploading OpenAPI files with remote and url reference pointers. 

On Main:
Screen Shot 2021-07-26 at 5 46 53 PM

With this PR:
Screen Shot 2021-07-26 at 5 47 18 PM

## 🧬 QA & Testing

- [x] Make sure that OpenAPI files with remote references are uploaded properly.


erunion (changes requested):
Can you add a test for this to make sure that what's being bundled and sent to the API is what we want?

__tests__/cmds/openapi.test.js
```diff
@@ -200,6 +200,41 @@ describe('rdme openapi', () => {
...
+      );
```
erunion, line 227: Can make this a little cleaner with snapshots:

```suggestion
      expect(requestBody).toMatchSnapshot();
```

src/cmds/openapi.js
```diff
@@ -98,9 +98,25 @@ exports.run = async function (opts) {
...
+        })
+        .catch(err => {
+          return Promise.reject(err);
+        });
+    }
+
     const options = {
       formData: {
-        spec: fs.createReadStream(path.resolve(process.cwd(), specPath)),
+        spec: bundledSpec || fs.createReadStream(path.resolve(process.cwd(), specPath)),
```
erunion, line 119: I forget what `specPath` is but when would that be present and `spec` (and `bundledSpec`) isn't?

__tests__/cmds/openapi.test.js
rahulhegdee: Dang, this is super convenient!

src/cmds/openapi.js
rahulhegdee: Thank you for catching this! I had meant to make the switch from spec to specPath before pushing the initial changes. We should be validating and bundling with specPath instead of spec.

erunion (approved):

PR #343, opened 2021-07-27
feat: making the docs command recursive by default

erunion:
## 🧰 Changes

* [x] Slightly alters the `--recursive` argument introduced in #313 to make the `docs` command recursive by default, always.
* [x] Updating our `docs` test to test recursive doc directory syncing.
* [x] Some general tidying up of the `docs` command to remove things that might look like internal IDs or API keys.

## 🧬 QA & Testing

* [ ] Tests pass?

### DIFF
--- a/__tests__/cmds/docs.test.js
+++ b/__tests__/cmds/docs.test.js
@@ -11,6 +11,19 @@ const docsEdit = require('../../src/cmds/docs/edit');
 const fixturesDir = `${__dirname}./../__fixtures__`;
 const key = 'Xmw4bGctRVIQz7R7dQXqH9nQe5d0SPQs';
 const version = '1.0.0';
+const category = '5ae9ece93a685f47efb9a97c';
+
+function getNockForVersion(v) {
+  return nock(config.host, {
+    reqheaders: {
+      'x-readme-version': v,
+    },
+  });
+}
+
+function hashFileContents(contents) {
+  return crypto.createHash('sha1').update(contents).digest('hex');
+}
 
 describe('rdme docs', () => {
   beforeAll(() => nock.disableNetConnect());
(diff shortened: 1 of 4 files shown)


__tests__/cmds/docs.test.js
```diff
@@ -39,64 +53,84 @@ describe('rdme docs', () => {
...
+        hash: hashFileContents(fileContents),
+      };
+
+      fileContents = fs.readFileSync(path.join(fixturesDir, `/existing-docs/subdir/another-doc.md`));
+      anotherDoc = {
+        slug: 'another-doc',
+        doc: frontMatter(fileContents),
+        hash: hashFileContents(fileContents),
+      };
+    });
```
rahulhegdee, line 73: love this change!

src/cmds/docs/index.js
```diff
@@ -119,7 +109,7 @@ exports.run = async function (opts) {
 
   const updatedDocs = await Promise.allSettled(
     files.map(async filename => {
-      const file = await readFile(path.join(folder, filename), 'utf8');
+      const file = await readFile(filename, 'utf8');
```
rahulhegdee, line 112: Seems like this change is breaking the command functionality. I believe we need path.join(folder, filename) to read the file from the correct path.

rahulhegdee (changes requested):
Needs a small change to bring back functionality, but otherwise LGTM!

src/cmds/docs/index.js
erunion: So I thought the same, but `files` are already being composed of fully pathed out files and this line was rewriting `path/to/doc.md` as `path/to/path/to/doc.md`.

rahulhegdee: You're right that files is fully pathed out initially but we reassign it to allFiles later on. And we stripped out the full paths in allFiles. 

With these lines:

`
// Pull off the leading subdirectory, to keep things consistant with below
  const allFiles = readdirRecursive(folder).map(file => file.replace(`${folder}${path.sep}`, ''));
`
`
  // Strip out non-markdown files
  const files = allFiles.filter(file => file.endsWith('.md') || file.endsWith('.markdown'));
`

rahulhegdee: So when we run updatedDocs, we don't have the fully pathed out filenames anymore and need to add back the folder lol

erunion: Talked offline and it's failing when you pass a `../relative/dir` path to the `folder` arg. `./dir` paths work just fine. Working on a fix.

erunion: Fixed in https://github.com/readmeio/rdme/pull/343/commits/c95aac8e3488e15b48919c8654edcf462b0c6592

rahulhegdee (approved):

PR #344, opened 2021-07-29
feat: add version selection to all relevant commands

rahulhegdee:
## 🧰 Changes

We had the option to select an existing version or create a new version with the openapi command if no version argument was provided, this PR adds that to all the commands that require the version argument.
- Adds version selection to the relevant docs and version commands
- Allows for specifying whether the user should receive the option to create a new version or just gives them a list of their current versions to choose from.

## 🧬 QA & Testing

- [x] Make sure tests pass. 
- [x] Make sure you can still input a version as a command line argument

### DIFF
--- a/__tests__/cmds/docs.test.js
+++ b/__tests__/cmds/docs.test.js
@@ -36,11 +36,6 @@ describe('rdme docs', () => {
     return expect(docs.run({})).rejects.toThrow('No project API key provided. Please use `--key`.');
   });
 
-  it('should error if no version provided', () => {
-    expect.assertions(1);
-    return expect(docs.run({ key })).rejects.toThrow('No project version provided. Please use `--version`.');
-  });
-
   it('should error if no folder provided', () => {
     expect.assertions(1);
     return expect(docs.run({ key, version: '1.0.0' })).rejects.toThrow(
@@ -103,6 +98,11 @@ describe('rdme docs', () => {
         .basicAuth({ user: key })
         .reply(200);
 
+      const versionMock = nock(config.host)
+        .get(`/api/v1/version/${version}`)
+        .basicAuth({ user: key })
+        .reply(200, { version });
+
       return docs.run({ folder: './__tests__/__fixtures__/existing-docs', key, version }).then(skippedDocs => {
         // All docs should have been updated because their hashes from the GET request were different from what they
         // are currently.
@@ -110,6 +110,7 @@ describe('rdme docs', () => {
 
         getMocks.done();
         updateMocks.done();
+        versionMock.done();
       });
     });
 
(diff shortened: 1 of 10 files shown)


src/lib/versionSelect.js
```diff
@@ -0,0 +1,38 @@
+const { prompt } = require('enquirer');
+const promptOpts = require('./prompts');
+const request = require('request-promise-native');
+const config = require('config');
+const APIError = require('./apiError');
+
+async function getSwaggerVersion(versionFlag, key, allowNewVersion) {
```
erunion, line 7: The versions API isn't specific to Swagger/OAS so we should rename this

```suggestion
async function getProjectVersion(versionFlag, key, allowNewVersion) {
```

src/lib/versionSelect.js
```diff
@@ -0,0 +1,38 @@
+const { prompt } = require('enquirer');
+const promptOpts = require('./prompts');
+const request = require('request-promise-native');
```
kanadgupta, line 3: Could we use [`node-fetch`](https://www.npmjs.com/package/node-fetch) here instead?

kanadgupta (commented):
This feature and the code overall looks great, excited for this! Just one small piece of feedback on our usage of `request-promise-native` 👀

src/lib/versionSelect.js
rahulhegdee: Should I update this for all the commands as well?

kanadgupta: Nah, not in this PR at least! I believe we already have a ticket for refactoring the other usages

erunion: Agreed that we should definitely move this to `node-fetch` so it's consistent with all of our OSS projects.

__tests__/cmds/docs.test.js
```diff
@@ -359,7 +392,12 @@ describe('rdme docs:edit', () => {
     const slug = 'getting-started';
     const body = 'abcdef';
 
-    nock(config.host).get(`/api/v1/docs/${slug}`).reply(200, { body });
+    nock(config.host)
```
erunion, line 395: Does this mock need to get cleaned up with `.done()` in the `.catch()`?

src/lib/versionSelect.js
rahulhegdee: Currently working on transitioning `request-promise-native` for all rdme commands to `node-fetch`. I'll have that in a separate PR!

erunion (approved):
this works great!

__tests__/cmds/docs.test.js
rahulhegdee: Oh interesting, yes it does! Will fix 👍.

PR #378, opened 2021-10-20
feat: adding support for OpenAPI 3.1

erunion:
## 🧰 Changes

* [x] Adds support for OpenAPI 3.1 during API definition validation.
* [x] Supports OpenAPI 3.1 definitions to be uploaded to projects.
* [x] Updates `oas` to support OpenAPI 3.1 within its `rdme oas generate` command for generating API definitions.

Fixes RM-1813

## 🧬 QA & Testing

See tests.


__tests__/cmds/openapi.test.js
```diff
@@ -68,36 +68,7 @@ describe('rdme openapi', () => {
     });
   });
 
-  it('should error if API errors', () => {
```
erunion, line 71: There's a whole bunch of nits like this in this changeset but the gist of what I did in this file are:

* Move errors into an "error handling" describe.
* Fix some confusing `it` language.
* Add some tests to make sure that errors that are coming back from `oas-normalize` (and `openapi-parser`) are of the new format.

src/cmds/openapi.js
```diff
@@ -105,8 +105,8 @@ exports.run = async function (opts) {
     }
 
     let bundledSpec;
-    const oas = new OASNormalize(specPath, { enablePaths: true });
-    await oas.validate().catch(err => {
+    const oas = new OASNormalize(specPath, { colorizeErrors: true, enablePaths: true });
+    await oas.validate(false, { validate: { colorizeErrors: true } }).catch(err => {
```
erunion, line 109: This option controls the pretty errors that'll be shown. If `colorizeErrors` is false (it's false by default) those errors will still look nice they just won't have colors or bold text.

julshotal (approved):
tests look good 🎉

Dashron (approved):
tested locally and it seems to be failing at readme.io, instead of in the lib, so it seems good to go once we're ready

PR #396, opened 2021-12-02
feat: swapping out the colors library for chalk

erunion:
## 🧰 Changes

This swaps out the [colors](https://npm.im/colors) library for [chalk](https://npm.im/chalk). `chalk` is a drop-in replacement for `colors`.

Why? 

1. `chalk` has a lot more dev support than `colors`.
2. I've always loathed how `colors` works by modifying `String.prototype`. Sure you can load `colors/safe` to do `colors.green()` but that's not the default flow.
3. `oas-normalize` and its dependency on `@readme/openapi-parser` uses it by way of `@readme/better-ajv-parser` so loading in `colors` here for yet another way of colorizing CLI messaging is silly.
4. We've been inconsistent in how we load it in the files that do have colors, so often you'll see `text.green` and it only works because that command is loaded in `./bin/rdme` where `require('colors')` was present. Woof.
5. It's got better detection for managing environments where CLI coloring might not be available.

## 🧬 QA & Testing

* [ ] All tests still pass?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,11 +25,15 @@ updates:
       prefix: chore(deps)
       prefix-development: chore(deps-dev)
     ignore:
+      # chalk is now an ESM package and can't be used here without a rewrite.
+      - dependency-name: chalk
+        versions:
+          - ">= 5"
       # node-fetch is now an ESM package and can't be used here without a rewrite.
       - dependency-name: node-fetch
         versions:
           - ">= 3"
-      # node-fetch is now an ESM package and can't be used here without a rewrite.
+      # table-layout is now an ESM package and can't be used here without a rewrite.
       - dependency-name: table-layout
         versions:
           - ">= 3"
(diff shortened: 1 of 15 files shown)


.github/dependabot.yml
```diff
@@ -25,11 +25,15 @@ updates:
       prefix: chore(deps)
       prefix-development: chore(deps-dev)
     ignore:
+      # chalk is now an ESM package and can't be used here without a rewrite.
```
RyanGWU82, line 28: Do we have a backlog item recorded anywhere to upgrade key projects so we can use ESM packages? It's going to be pretty painful if we get stuck on old versions of all our dependencies.

RyanGWU82 (approved):
LGTM! 🎉

.github/dependabot.yml
erunion: Agreed. We don't have any backlog items for this and we haven't really talked about it at all. It's going to be really painful to upgrade everything because it's going to require us to go ESM across **everything** at the same time.

RyanGWU82: I'll file a ticket about this and make sure it's on Dom's radar -- sounds like painful tech debt but something we'll need to deal with sooner or later.

Issue #401, opened 2021-12-07
6.0.0 won't install on the `node:lts-alpine` image


magni-:
We sync our API docs on `node:lts-alpine` images via CircleCI. This has worked fine since we first started doing it in late September (meaning using `rdme` versions 5.1.4 and 5.2.0), but now blows up with the new 6.0.0 release.

The error message from `npm` is rather cryptic:
```
npm ERR! syscall spawn git
npm ERR! path git
npm ERR! errno -2
npm ERR! enoent An unknown git error occurred
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```

Here are the full logs for the failed 6.0.0 install, along with logs from a subsequent successful install of 5.2.0 in the same VM.

[npm-install-g-5-2-0.log](https://github.com/readmeio/rdme/files/7666045/npm-install-g-5-2-0.log)
[npm-install-g-6-0-0.log](https://github.com/readmeio/rdme/files/7666046/npm-install-g-6-0-0.log)

I tried taking a look at both of these to see if there was anything obvious but couldn't spot it. `timing idealTree Completed` shows up in the 5.2.0 log but not the 6.0.0 log. I don't know what that line represents 🤷🏼

erunion:
Looks like the 6.0 install is using npm@8.1.2 but the 5.2 one was using npm@8.2. Do you mind retrying the 6.0 install with npm@8.2 and sending me the log from that if it fails again?

magni-:
Ah yep, I noticed `npm` wasn't at the latest and tried updating that to see if it fixed it—it didn't. Here's a full log:
[npm-install-g-6-0-0.log](https://github.com/readmeio/rdme/files/7672290/npm-install-g-6-0-0.log)

erunion:
Seems this is happening because rdme depends on [oas](https://npm.im/oas), which depends on [@readme/openapi-parser](https://npm.im/@readme/openapi-parser) and that's loading in a fork we have of [@apidevtools/json-schema-ref-parser](https://npm.im/@apidevtools/json-schema-ref-parser) and though that dependency there is set to `github:erunion/json-schema-ref-parser#fix/dates-as-strings` npm is loading it with git+ssh and the `node:lts-alpine` image doesn't have git installed.

I'll publish that fork to npm like a proper package and publish a 6.0.1 release of rdme but in the meantime this image was able to install rdme:

```dockerfile
FROM node:lts-alpine

ADD . /src
WORKDIR /src

[...]

PR #402, opened 2021-12-08
fix: upgrading oas and oas-normalize to fix an npm installation issue

erunion:
## 🧰 Changes

Upgrades `oas` and `oas-normalize` to fix an installation issue with one of the dependencies of `oas-normalize` where because it was in the lock file as `git+ssh://` you needed Git available on the machine to install `rdme`.

This fixes https://github.com/readmeio/rdme/issues/401.

## 🧬 QA & Testing

There's no `git+ssh://` instances in `package-lock.json` anymore so if all tests still pass then this is good.


package-lock.json
```diff
@@ -1537,21 +1514,10 @@
         "openapi-types": ">=7"
       }
     },
-    "node_modules/@readme/openapi-parser/node_modules/@apidevtools/json-schema-ref-parser": {
-      "version": "0.0.0-dev",
-      "resolved": "git+ssh://git@github.com/erunion/json-schema-ref-parser.git#402904c92c8465db788be70655c3773c6e42a051",
```
erunion, line 1542: This is the dependency that was breaking installation.

domharrington (approved):

PR #406, opened 2021-12-15
docs: add/improve our command warnings

kanadgupta:
## 🧰 Changes

- [x] Added a warning for when someone uses the `openapi` command with both a `version` and `id` parameter, since the `version` parameter is ignored in this case. This has created a lot of confusion for users so hopefully this warning should provide some clarification
   - [x] Added a test for this
- [x] Enhanced language and formatting in several warnings 🟡 (one potential enhancement idea: what if we had an ESLint rule or something that prevented us from using `console.warn` and instead we had to use a standard shared chalk-formatted logger function? cc: @erunion)
- [x] Updated a test to test against our `openapi` command and refreshed its corresponding snapshot (see [this comment](https://github.com/readmeio/rdme/pull/406#discussion_r770807518))

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -69,15 +69,15 @@ describe('cli', () => {
     });
 
     it('should print usage for a given command', async () => {
-      await expect(cli(['swagger', '--help'])).resolves.toMatchSnapshot();
+      await expect(cli(['openapi', '--help'])).resolves.toMatchSnapshot();
     });
 
     it('should print usage for a given command if supplied as `help <command>`', async () => {
-      await expect(cli(['help', 'swagger'])).resolves.toMatchSnapshot();
+      await expect(cli(['help', 'openapi'])).resolves.toMatchSnapshot();
     });
 
     it('should not surface args that are designated as hidden', async () => {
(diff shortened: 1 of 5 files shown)


__tests__/index.test.js
```diff
@@ -69,15 +69,15 @@ describe('cli', () => {
...
     });
 
     it('should print usage for a given command if supplied as `help <command>`', async () => {
-      await expect(cli(['help', 'swagger'])).resolves.toMatchSnapshot();
+      await expect(cli(['help', 'openapi'])).resolves.toMatchSnapshot();
     });
 
     it('should not surface args that are designated as hidden', async () => {
-      await expect(cli(['swagger', '--help'])).resolves.toMatchSnapshot();
+      await expect(cli(['openapi', '--help'])).resolves.toMatchSnapshot();
```
kanadgupta, line 80: I figured it makes more sense to test our `openapi` command rather than our deprecated `swagger` command here, which resulted in a bunch of snapshot updates in `__tests__/__snapshots__/index.test.js.snap`.

erunion (approved):

PR #409, opened 2021-12-17
fix(openapi): return rejected `Promise` if spec uploads fail

### LINKED ISSUE #407: Non-zero exit status on failure
It would be great to have a non-zero exit status if a command fails (specifically `rdme openapi`) that way our ci will fail the build instead of reporting success when it wasn't successful.

Thanks for all the hard work!

kanadgupta:
## 🧰 Changes

Turns out we had an edge case where we were returning a `console.log()` rather than a rejected `Promise` when we received errors from the API, resulting in an incorrectly successful exit code. This PR fixes that in https://github.com/readmeio/rdme/pull/409/commits/0f2372ed1fc0b94ddba546a5a162b35297c0444b so we properly return an exit code of 1. And as expected, some housekeeping:
- [x] Refactors existing tests to check for the rejected `Promise` rather than a console output https://github.com/readmeio/rdme/commit/8811c8ce6945e33958d7979668e95d39e85e3f54
- [x] Adds additional test coverage to check for our other API error edge cases. We're now over 90% line coverage in the OpenAPI command file! 📈 https://github.com/readmeio/rdme/commit/8811c8ce6945e33958d7979668e95d39e85e3f54
- [x] Slightly reformats our rejected `Promise` error outputs to be padded with newlines https://github.com/readmeio/rdme/commit/9936de5d68e6c9c77aebc762001d4076ff871115
- [x] Small fix to our general command handler so it properly uses our lengthier error message for invalid commands. https://github.com/readmeio/rdme/commit/3b75d297c32813a4fcd69dba009d92ff4e4439b1 See below for a before and after:

image

Resolves #407.

## 🧬 QA & Testing

If the tests pass, we should be good! But as an extra check, you can check out this branch and run the following commands to confirm that we see a non-zero exit code:

```sh
$ bin/rdme openapi --id="asdf" --key="asdf" ./__tests__/__fixtures__/ref-oas/petstore.json

We couldn't find your API key.

If you need help, email support@readme.io and include the following link to your API log: 'https://docs.readme.com/logs/xxx-xxx'.

$ echo $?
1
```

### DIFF
--- a/__tests__/cmds/openapi.test.js
+++ b/__tests__/cmds/openapi.test.js
@@ -4,6 +4,7 @@ const fs = require('fs');
 const promptHandler = require('../../src/lib/prompts');
 const swagger = require('../../src/cmds/swagger');
 const openapi = require('../../src/cmds/openapi');
+const APIError = require('../../src/lib/apiError');
 
 const key = 'API_KEY';
 const id = '5aa0409b7cf527a93bfb44df';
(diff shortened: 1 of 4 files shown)


erunion (approved):
can't believe we weren't throwing an exit(1) for failed uploads 😞

PR #424, opened 2022-01-09
refactor: commands into command classes

erunion:
## 🧰 Changes

> ⚠️ There's a lot of whitespace changes in this PR so make sure to view it with that hidden.

While thinking around with Typescript here this evening I started migrating every command into being a stricter class that implements a `Command` interface (in order for Typescript to understand each command).

Since I quite liked how the commands felt while nestled inside of a more stricter class environment I've reworked that work here so that every command has a constructor where it sets up its arguments and `--help` screen docs, and an async `run` method to run the command.

If I manage to get the Typescript work reviewable it'll make that PR a lot easier to read with this refactoring out of the way.

## 🧬 QA & Testing

* [ ] Do you like these changes?
* [ ] Tests still pass?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,8 @@
   "extends": ["@readme/eslint-config"],
   "root": true,
   "rules": {
+    "class-methods-use-this": ["error", { "exceptMethods": ["run"] }],
+
     /**
      * This is a small rule to prevent us from using console.log() statements in our commands.
      *
--- a/__tests__/cmds/versions.test.js
+++ b/__tests__/cmds/versions.test.js
@@ -2,10 +2,10 @@ const nock = require('nock');
 const config = require('config');
 const promptHandler = require('../../src/lib/prompts');
 
-const versions = require('../../src/cmds/versions');
-const createVersion = require('../../src/cmds/versions/create');
-const deleteVersion = require('../../src/cmds/versions/delete');
-const updateVersion = require('../../src/cmds/versions/update');
+const VersionsCommand = require('../../src/cmds/versions');
+const CreateVersionCommand = require('../../src/cmds/versions/create');
+const DeleteVersionCommand = require('../../src/cmds/versions/delete');
+const UpdateVersionCommand = require('../../src/cmds/versions/update');
 
 const key = 'API_KEY';
 const version = '1.0.0';
@@ -39,6 +39,8 @@ describe('rdme versions*', () => {
   afterEach(() => nock.cleanAll());
 
   describe('rdme versions', () => {
+    const versions = new VersionsCommand();
+
     it('should error if no api key provided', async () => {
       await expect(versions.run({})).rejects.toThrow('No project API key provided. Please use `--key`.');
     });
@@ -91,6 +93,8 @@ describe('rdme versions*', () => {
   });
 
   describe('rdme versions:create', () => {
+    const createVersion = new CreateVersionCommand();
+
     it('should error if no api key provided', () => {
       expect.assertions(1);
       return createVersion.run({}).catch(err => {
@@ -140,6 +144,8 @@ describe('rdme versions*', () => {
   });
 
   describe('rdme versions:delete', () => {
+    const deleteVersion = new DeleteVersionCommand();
+
     it('should error if no api key provided', async () => {
       await expect(deleteVersion.run({})).rejects.toThrow('No project API key provided. Please use `--key`.');
     });
@@ -178,6 +184,8 @@ describe('rdme versions*', () => {
   });
 
   describe('rdme versions:update', () => {
+    const updateVersion = new UpdateVersionCommand();
+
     it('should error if no api key provided', async () => {
       await expect(updateVersion.run({})).rejects.toThrow('No project API key provided. Please use `--key`.');
     });
--- a/src/cmds/login.js
+++ b/src/cmds/login.js
@@ -9,68 +9,72 @@ const fetch = require('node-fetch');
 
 const testing = process.env.NODE_ENV === 'testing';
 
-exports.command = 'login';
-exports.usage = 'login [options]';
-exports.description = 'Login to a ReadMe project.';
-exports.category = 'admin';
-exports.position = 1;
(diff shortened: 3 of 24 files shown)


src/cmds/versions/create.js
```diff
@@ -2,104 +2,104 @@ const config = require('config');
...
+      body: JSON.stringify({
+        version,
+        codename: codename || '',
+        is_stable: main === 'true' || promptResponse.is_stable,
+        is_beta: beta === 'true' || promptResponse.is_beta,
+        from: fork || promptResponse.from,
+        is_hidden: promptResponse.is_stable ? false : !(isPublic === 'true' || promptResponse.is_hidden),
+      }),
+    })
+      .then(handleRes)
```
erunion, line 100: Moved this over to use `handleRes` because we can here.

src/cmds/whoami.js
```diff
@@ -1,24 +1,27 @@
...
+    this.usage = 'whoami';
+    this.description = 'Displays the current user and project authenticated with ReadMe.';
+    this.category = 'admin';
+    this.position = 3;
 
-exports.args = [];
-
-exports.run = () => {
-  if (!configStore.has('email') || !configStore.has('project')) {
-    return Promise.reject(new Error(`Please login using \`${config.get('cli')} ${loginCmd.command}\`.`));
```
erunion, line 16: I stopped directly calling the login command for its name here because we know it. It's login.

src/cmds/login.js
```diff
@@ -9,68 +9,72 @@ const fetch = require('node-fetch');
...
-    token: opts['2fa'] && (await read({ prompt: '2fa token:' })),
-  };
-}
+  async run(opts) {
+    let { email, password, project, token } = opts;
 
-exports.run = async function (opts) {
-  let { email, password, project, token } = opts;
+    /* istanbul ignore next */
+    async function getCredentials() {
```
erunion, line 38: Since I've moved this inside of `run` we don't need to pass it `opts` anymore.

src/cmds/versions/index.js
```diff
@@ -1,125 +1,129 @@
...
+            new Error(
+              `Sorry, you haven't created any versions yet! See \`${config.get('cli')} help ${
+                new CreateVersionCmd().command
+              }\` for commands on how to do that.`
+            )
+          );
+        }
+
+        if (version === undefined) {
+          return Promise.resolve(VersionsCommand.getVersionsAsTable(versions));
```
erunion, line 123: Since `getVersionsAsTable` is a static method we need to call it off the main class, `this.getVersionsAsTable` won't work. The method doesn't really need to be static, but since it doesn't use any internal command data it probably makes sense for it to be static?

If not, we'll need to alias `this` to `self` at the top of `run` to get around scope issues within the `fetch` thenable that this line is at.

src/cmds/swagger.js
```diff
@@ -1,16 +1,18 @@
...
+const OpenAPICommand = require('./openapi');
 
-exports.command = 'swagger';
-exports.usage = 'swagger [file] [options]';
-exports.description = 'Alias for `rdme openapi`. [deprecated]';
-exports.category = openapi.category;
-exports.position = openapi.position + 1;
+module.exports = class SwaggerCommand extends OpenAPICommand {
+  constructor() {
+    super();
```
kanadgupta, line 6: Much cleaner 🎉

src/cmds/versions/create.js
kanadgupta: niiiice

.eslintrc
```diff
@@ -2,6 +2,8 @@
   "extends": ["@readme/eslint-config"],
   "root": true,
   "rules": {
+    "class-methods-use-this": ["error", { "exceptMethods": ["run"] }],
```
kanadgupta, line 5: Wanna add a small comment explaining why we need this?

kanadgupta (approved):
I generally don't love using classes in JS if I can avoid it, but it makes sense with some of our command patterns and cleans things up nicely in certain ways. LGTM!

src/lib/commands.js
```diff
@@ -54,11 +55,11 @@ exports.list = () => {
 
   files.forEach(file => {
     // eslint-disable-next-line global-require, import/no-dynamic-require
-    const command = require(file);
+    const Command = require(file);
 
     commands.push({
       file,
-      command,
+      command: new Command(),
```
Dashron, line 62: Thoughts on `new` here and load vs. `module.exports = new (commandClass)();` in each file?

I've used both techniques in the past. `new` here and load is nice to ensure that we don't have side effects leaking between calls, and might be what I prefer, but I figured I'd bring it up if you thought it would be useful.

Dashron (approved):

Dashron:
I def like the classes, clean and allows for interesting, clean improvements like inheritance.

src/lib/commands.js
erunion: I think I prefer to do the instantiation on load instead of in the export.

__tests__/cmds/versions.test.js
```diff
@@ -52,7 +54,9 @@ describe('rdme versions*', () => {
         .basicAuth({ user: key })
         .reply(200, [versionPayload, version2Payload]);
 
-      await expect(versions.run({ key })).resolves.toMatchSnapshot();
+      const table = await versions.run({ key });
```
erunion, line 57: Reverted this back to using these two assertions because the `FORCE_COLOR` env variable doesn't seem to always get used between `colors` and `chalk`.

kanadgupta (approved):

kanadgupta:
Also wait we probably don't need both of these right :eyes:

https://github.com/readmeio/rdme/blob/2bd4199da76c723f5c6f49baccf76d97d76f0b56/.github/workflows/ci.yml#L25-L28
https://github.com/readmeio/rdme/blob/2bd4199da76c723f5c6f49baccf76d97d76f0b56/__tests__/set-node-env.js#L1-L3

erunion:
Ah good eye I'll remove the one in the workflow config.

PR #435, opened 2022-02-02
feat: add update-notifier to warn when using an old version of rdme

domharrington:
## 🧰 Changes
Add update-notifier to warn when using an old version of rdme

Example of what it looks like:
image

## 🧬 QA & Testing
To test this out, you can replace the line in bin/rdme with the following:

```
updateNotifier({ pkg: { name: 'rdme', version: '1.0.0' }, updateCheckInterval: 0 }).notify();
```

Example nabbed from here:
https://github.com/yeoman/update-notifier/blob/HEAD/example.js

### DIFF
--- a/bin/rdme
+++ b/bin/rdme
@@ -1,6 +1,11 @@
 #! /usr/bin/env node
 const chalk = require('chalk');
 
+const updateNotifier = require('update-notifier');
+const pkg = require('../package.json');
+
+updateNotifier({ pkg }).notify();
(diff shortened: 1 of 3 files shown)


erunion (approved):
works great locally

kanadgupta (approved):
Ohhh I knew I had seen this before... the npm CLI itself actually uses this, nice! LGTM 👍🏽

PR #436, opened 2022-02-02
feat: sending a custom rdme user agent with all requests

erunion:
## 🧰 Changes

Much to my surprise and chagrin we weren't already sending a custom `rdme/(major).(minor).(patch)` user agent with all API requests. This fixes that!

## 🧬 QA & Testing

I've slightly shuffled around the `cleanHeaders` and `handleRes` libraries we had in `src/lib` into this new `fetch` wrapper in `src/lib/fetch`. Anything that once loaded `node-fetch` now loads this (which in turn loads `node-fetch`).

I've also added some unit tests for this to make sure that user agents are getting sent, changed every command test that mocks calls to our API to mock that API call + the user agent, and also cleaned up a few command tests that didn't have any explicit assertions (they were just silently testing either the mocks would succeed or that the command wouldn't fail).


__tests__/.eslintrc
```diff
@@ -1,20 +1,3 @@
 {
-  "extends": ["@readme/eslint-config/testing"],
-  "env": {
-    "jest": true
-  },
-  "rules": {
-    "jest/expect-expect": [
```
erunion, line 7: All tests have explicit `expect()`s now so we don't need add these exclusions for nock.

__tests__/cmds/docs.test.js
```diff
@@ -178,16 +181,24 @@ describe('rdme docs', () => {
...
-      });
+      await expect(docs.run({ folder: './__tests__/__fixtures__/new-docs', key, version })).resolves.toStrictEqual([
+        {
+          slug: 'new-doc',
+          body: '\nBody\n',
+          category: '5ae122e10fdf4e39bb34db6f',
+          title: 'This is the document title',
+          lastUpdatedHash: 'a23046c1e9d8ab47f8875ae7c5e429cb95be1c48',
+        },
+      ]);
```
kanadgupta, line 197: ugh we really had these tests in so many places lol 😭

__tests__/cmds/docs.test.js
```diff
@@ -285,21 +296,29 @@ describe('rdme docs', () => {
...
         .reply(200, { version });
 
-      return docs.run({ folder: './__tests__/__fixtures__/slug-docs', key, version }).then(() => {
-        getMock.done();
-        postMock.done();
-        versionMock.done();
-      });
+      await expect(docs.run({ folder: './__tests__/__fixtures__/slug-docs', key, version })).resolves.toStrictEqual([
+        {
+          slug: 'marc-actually-wrote-a-test',
```
kanadgupta, line 311: lmao @mjcuva

__tests__/lib/commands.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-conditional-expect */
```
kanadgupta, line 1: Could we disable it for each line or is it affecting the whole file?

src/lib/fetch.js
```diff
@@ -1,10 +1,42 @@
...
+      'User-Agent': userAgent,
+    };
+  } else {
+    options.headers['User-Agent'] = userAgent;
+  }
+
+  return fetch(url, options);
+};
+
+module.exports.handleRes = async function handleRes(res) {
```
kanadgupta, line 24: Could you add a li'l description comment above this function for consistency's sake?

kanadgupta (approved):
A couple of small notes but otherwise LGTM! The refactor is real nice 👏🏽 

Also quite bummed we hadn't been adding the user agent until now 😞

__tests__/lib/commands.test.js
erunion: Every test in this one has a conditional expect unfortunately.

PR #437, opened 2022-02-11
feat: add wrapper for GitHub Actions

kanadgupta:
## 🧰 Changes

This finally does what we've been talking about for months and adds an extremely lightweight wrapper around this repo so we can have first-class support for GitHub Actions 🚀 

You can check out [`action.yml`](https://github.com/readmeio/rdme/blob/deprecate-our-github-action-rm-1944/action.yml) for all the heavy lifting.

Here's how uploading an OpenAPI file looks with `npx` vs. the GitHub Action:
```yml
# npx
- run: npx -y rdme@6 openapi openapi.json --key=API_KEY --id=API_SETTING_ID
```
```yml
# GitHub Action (🆕!)
- uses: readmeio/rdme@6
  with:
    rdme: openapi openapi.json --key=API_KEY --id=API_SETTING_ID
```

<details>
<summary>And here's what it looks like with the old GitHub Action:</summary>

```yml
# GitHub Action (legacy)
- uses: readmeio/github-readme-sync@v2
  with:
    readme-oas-key: 'API_KEY:API_SETTING_ID'
    oas-file-path: openapi.json
```

</details>

I think I prefer the `npx` notation to be honest, but there are a few benefits of this new wrapper we've created:
- Users can [configure our good friend dependabot to keep their version up-to-date](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot) (and I don't _think_ this is possible with inline `run: npx` commands in your GitHub workflow files, unless they list `rdme` in their dev-deps)
- We can position `rdme` as a first-class GitHub Actions integration, rather than a workaround, meaning we can...
  - list this in [the GitHub Actions Marketplace](https://github.com/marketplace?type=actions) 🛍️ 
  - safely destroy [`[redacted]`](https://github.com/readmeio/github-readme-sync) once and for all 🚮 

## 🧬 QA & Testing

[...]

PR #440, opened 2022-02-16
feat: dropping support for node 12

erunion:
> ⚠️ Do not merge this until the new Github Action workflow has been tagged to the current v6 release series.

## 🧰 Changes

Node 12 hits EOL in April and with our new action in #437 it'd be nice to start with a clean start, breaking change for folks, so we might as well drop support for Node 12 now and publish a breaking change rather than having to publish another breaking change in a few months. 

This also allows us to _finally_ start using optional chaining in this codebase as that was introduced in Node 14.

## 🧬 QA & Testing

* [ ] All tests still passing?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,9 @@
 {
   "extends": ["@readme/eslint-config"],
   "root": true,
+  "parserOptions": {
+    "ecmaVersion": 2020
+  },
   "rules": {
     /**
      * Because our command classes have a `run` method that might not always call `this` we need to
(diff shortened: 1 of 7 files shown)


kanadgupta (commented):
Found a few other places we could swap out:
https://github.com/readmeio/rdme/blob/a5e8fdb5000df513b697008b9251fa1a55ac14da/bin/rdme#L20
https://github.com/readmeio/rdme/blob/a5e8fdb5000df513b697008b9251fa1a55ac14da/src/lib/prompts.js#L151

kanadgupta (approved):
Much cleaner 🎉

PR #441, opened 2022-02-16
docs: update Action metadata

kanadgupta:
Seeing this error when publishing to the Marketplace:
image

Worked on these edits with Sarah!

Fixes RM-3574

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,9 +1,9 @@
-name: rdme 🦉
+name: rdme 🦉 Sync to ReadMe
 author: ReadMe
 branding:
   color: blue
   icon: book-open
-description: Maintaining OpenAPI definitions or Markdown documentation on GitHub? Sync it to your ReadMe Developer Hubs!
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -1,9 +1,9 @@
-name: rdme
+name: rdme 🦉 Sync to ReadMe
 author: ReadMe
 branding:
   color: blue
   icon: book-open
-description: Update your ReadMe developer hubs in a CI environment.
+description: Sync OpenAPI (OAS) definitions and Markdown documentation from your GitHub repository to your ReadMe developer hub
```
erunion, line 6: no punctuation?

erunion (approved):

action.yml
kanadgupta: That seems to be the pattern most other listings follow 🤷🏽  https://github.com/marketplace?type=actions

PR #439, opened 2022-02-15
docs: add GitHub Actions guidance

kanadgupta:
This is a doozy. This PR updates our README.md, adds docs for [our knowledge base](https://docs.readme.com/docs/rdme), and sets up a CI workflow for doing a find/replace and syncing the knowledge base page to ReadMe.

Fixes RM-3553

PR #446, opened 2022-02-23
feat: debugging support

kanadgupta:
## 🧰 Changes

<details>
<summary><b>Commands that now have debugging</b></summary>

- [x] `login`
- [x] `logout`
- [x] `whoami`
- [x] `openapi`
- [x] `validate`
- [x] `swagger`
- [x] `docs`
- [x] `docs:edit`
- [x] `versions`
- [x] `versions:create`
- [x] `versions:update`
- [x] `versions:delete`
- [x] `oas`
- [x] `open`

</details>

<details>
<summary><b>Helper functions that now have debugging</b></summary>

- [x] `fetch()`
- [x] `handleRes()`

Obviously there are many other helper functions, but these are the only ones that felt necessary.

</details>

Additionally, I made a small enhancement to `bin/rdme` to output errors via the native GitHub Action annotation format so they're properly surfaced. You can see an example of this in [this workflow run](https://github.com/readmeio/rdme/actions/runs/1890909393) (and in the image below[^1]).

image

## 🧬 QA & Testing

No functional changes beyond migrating `isGHA()` into its own file and enhancing the logging in `bin/rdme` (the results of which you can see in the annotation example above). Do tests pass? And can you see debug logs on [this workflow run](https://github.com/readmeio/rdme/runs/5313514327?check_suite_focus=true#step:3:116)?

Fixes RM-3451, RM-3544

[^1]: Yes, the "Process completed with exit code 1" error is excessive and is bugging me as well, and no, I don't know how to get rid of it... I think this is an unfortunate by-product of going with the composite action route 😞

### DIFF
--- a/bin/rdme
+++ b/bin/rdme
@@ -36,6 +36,5 @@ require('../src')(process.argv.slice(2))
 
     // eslint-disable-next-line no-console
     console.error(chalk.red(`\n${message}\n`));
-
     return process.exit(1);
   });
--- a/bin/rdme
+++ b/bin/rdme
@@ -1,8 +1,11 @@
 #! /usr/bin/env node
 const chalk = require('chalk');
+const core = require('@actions/core');
 
 const updateNotifier = require('update-notifier');
 const pkg = require('../package.json');
+
+const isGHA = require('../src/lib/isGitHub');
 
 updateNotifier({ pkg }).notify();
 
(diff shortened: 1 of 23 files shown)


bin/rdme
```diff
@@ -21,6 +24,16 @@ require('../src')(process.argv.slice(2))
       message = err.message;
     }
 
+    /**
+     * If we're in a GitHub Actions environment,
+     * log errors with that formatting instead.
```
erunion, line 29:
```suggestion
     * If we're in a GitHub Actions environment, log errors with that formatting instead.
```

src/lib/isGitHub.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Small env check to determine if we're in a GitHub Actions environment
+ * @link https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+ */
+module.exports = function isGHA() {
+  return process.env.GITHUB_ACTIONS === 'true';
```
erunion, line 6: We can replace this with the [ci-info](https://npm.im/ci-info) library I'm pulling in for RM-3621 to do `require('ci-info').GITHUB_ACTIONS`. That checks the `GITHUB_ACTIONS` env variable and also has awareness if it's part of a pull request.

https://github.com/watson/ci-info/blob/master/vendors.json#L93-L96

erunion (changes requested):
Can you add a bit to the README about how to generate these debugging messages? Will be helpful for support folks to know about.

kanadgupta:
>Can you add a bit to the README about how to generate these debugging messages? Will be helpful for support folks to know about.

@erunion it's on my list of outstanding items for #439!

erunion (approved):

PR #449, opened 2022-02-24
feat: ci awareness to disable terminal prompting within commands

erunion:
| 🚥 Fix RM-3621 |
| :-- |

## 🧰 Changes

With our new workflow in #437 there's a new issue that will likely crop up for some folks is that if you don't supply all of the required arguments for a command in a CI environment, our [enquirer](https://npm.im/enquirer) setup will prompt the terminal for input -- something that will never be answered in a CI environment.

For this we're introducing some new awareness to our commands that ask the user for input to not do that if it's within a CI environment and instead throw an error telling them that they're missing arguments.

I'm also making some other adjustments to a couple commands to clean up their argument structure:

* [x] `versions:create`
  * `--main` and `--beta` are now booleans
  * `--isPublic` has been renamed to just `--public` and it's also been made into a boolean.
  * [ ] Make sure usage in README.md is updated
* [ ] `versions:update`
  * `--main` and `--beta` are now booleans
  * `--isPublic` has been renamed to just `--public` and it's also been made into a boolean.
  * [ ] Make sure usage in README.md is updated
## 🧬 QA & Testing

If with all the changes I made tests are still passing, 👍 

You can also test this behavior out locally by prefixing `GITHUB_ACTIONS=true` to a `./bin/rdme` call:

```
$ GITHUB_ACTIONS=true ./bin/rdme versions:create --version=1.0.0

We've detected you're running within a GitHub Actions environment, please supply the following required 
arguments for your command instead of relying on our terminal prompt system: fork
```

### DIFF
--- a/__tests__/cmds/versions.test.js
+++ b/__tests__/cmds/versions.test.js
@@ -155,9 +155,7 @@ describe('rdme versions*', () => {
           .reply(200, [{ version }, { version }]);
 
         await expect(createVersion.run({ key, version })).rejects.toThrow(
-          new Error(
-            "We've detected you're running within a CircleCI environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: fork"
-          )
+          /We've detected you're running within a (.*) environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: fork/
         );
 
         mockRequest.done();
--- a/__tests__/cmds/versions.test.js
+++ b/__tests__/cmds/versions.test.js
@@ -37,6 +37,8 @@ jest.mock('../../src/lib/prompts');
 describe('rdme versions*', () => {
   beforeAll(() => nock.disableNetConnect());
 
+  beforeEach(() => jest.resetModules());
+
   afterEach(() => nock.cleanAll());
 
   describe('rdme versions', () => {
@@ -103,30 +105,36 @@ describe('rdme versions*', () => {
     });
 
     it('should create a specific version', async () => {
-      promptHandler.createVersionPrompt.mockResolvedValue({
-        is_stable: true,
-        is_beta: false,
-        from: '1.0.0',
-      });
-
       const mockRequest = getApiNock()
-        .get('/api/v1/version')
-        .basicAuth({ user: key })
-        .reply(200, [{ version }, { version }])
-        .post('/api/v1/version')
+        .post('/api/v1/version', {
+          version: '1.0.0',
+          codename: '',
+          is_stable: false,
+          is_beta: false,
+          from: '1.0.0',
+          is_hidden: true,
+        })
         .basicAuth({ user: key })
         .reply(201, { version });
 
-      await expect(createVersion.run({ key, version })).resolves.toBe('Version 1.0.0 created successfully.');
+      await expect(createVersion.run({ key, version, fork: '1.0.0', main: true, beta: false })).resolves.toBe(
+        'Version 1.0.0 created successfully.'
+      );
+
       mockRequest.done();
     });
 
+    it.todo('should prompt a version list if no `--fork` argument is supplied');
+
     it('should catch any post request errors', async () => {
-      expect.assertions(1);
-      promptHandler.createVersionPrompt.mockResolvedValue({
-        is_stable: false,
-        is_beta: false,
-      });
+      const args = {
+        key,
+        version,
+        fork: '0.0.5',
+        main: false,
+        beta: false,
+        public: false,
+      };
 
       const errorResponse = {
         error: 'VERSION_EMPTY',
@@ -137,10 +145,33 @@ describe('rdme versions*', () => {
 
       const mockRequest = getApiNock().post('/api/v1/version').basicAuth({ user: key }).reply(400, errorResponse);
 
-      await expect(createVersion.run({ key, version, fork: '0.0.5' })).rejects.toStrictEqual(
-        new APIError(errorResponse)
-      );
+      await expect(createVersion.run(args)).rejects.toThrow(new APIError(errorResponse));
       mockRequest.done();
+    });
+
+    describe('ci environments', () => {
+      beforeEach(() => {
+        process.env.CIRCLECI = true;
+      });
+
+      afterEach(() => {
+        delete process.env.CIRCLECI;
+      });
+
+      it('should prompt for missing required arguments if in a CI environment', async () => {
+        const mockRequest = getApiNock()
+          .get('/api/v1/version')
+          .basicAuth({ user: key })
+          .reply(200, [{ version }, { version }]);
+
+        await expect(createVersion.run({ key, version })).rejects.toThrow(
+          new Error(
+            "We've detected you're running within a CircleCI environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: fork"
+          )
+        );
+
+        mockRequest.done();
+      });
     });
   });
 
--- a/__tests__/cmds/versions.test.js
+++ b/__tests__/cmds/versions.test.js
@@ -155,7 +155,7 @@ describe('rdme versions*', () => {
           .reply(200, [{ version }, { version }]);
 
         await expect(createVersion.run({ key, version })).rejects.toThrow(
-          /We've detected you're running within a (.*) environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: fork/
+          /we noticed you're running this command within/i
         );
 
         mockRequest.done();
--- a/__tests__/cmds/versions.test.js
+++ b/__tests__/cmds/versions.test.js
@@ -105,37 +105,27 @@ describe('rdme versions*', () => {
     });
 
     it('should create a specific version', async () => {
+      const args = { key, version, fork: '1.0.0', main: false, beta: true, public: true };
       const mockRequest = getApiNock()
         .post('/api/v1/version', {
           version: '1.0.0',
           codename: '',
           is_stable: false,
-          is_beta: false,
+          is_beta: true,
           from: '1.0.0',
-          is_hidden: true,
+          is_hidden: false,
         })
         .basicAuth({ user: key })
         .reply(201, { version });
 
-      await expect(createVersion.run({ key, version, fork: '1.0.0', main: true, beta: false })).resolves.toBe(
-        'Version 1.0.0 created successfully.'
-      );
-
+      await expect(createVersion.run(args)).resolves.toBe('Version 1.0.0 created successfully.');
       mockRequest.done();
     });
 
     it.todo('should prompt a version list if no `--fork` argument is supplied');
 
     it('should catch any post request errors', async () => {
-      const args = {
-        key,
-        version,
-        fork: '0.0.5',
-        main: false,
-        beta: false,
-        public: false,
-      };
-
+      const args = { key, version, fork: '0.0.5', main: false, beta: false, public: false };
       const errorResponse = {
         error: 'VERSION_EMPTY',
         message: 'You need to include an x-readme-version header',
--- a/src/cmds/versions/create.js
+++ b/src/cmds/versions/create.js
@@ -118,7 +118,7 @@ module.exports = class CreateVersionCommand {
         is_stable: opts.main || answers.main,
         is_beta: opts.beta || answers.beta,
         from: opts.fork || answers.fork,
-        is_hidden: answers.is_stable ? false : !('public' in opts || answers.public),
+        is_hidden: opts.main || answers.main ? false : !('public' in opts || answers.public),
       }),
     })
       .then(handleRes)
--- a/src/cmds/versions/create.js
+++ b/src/cmds/versions/create.js
@@ -1,6 +1,6 @@
 const config = require('config');
 const semver = require('semver');
-const enquirer = require('../../lib/enquirer');
+const prompt = require('../../lib/prompt');
 
 const fetch = require('../../lib/fetch');
 const { cleanHeaders, handleRes } = require('../../lib/fetch');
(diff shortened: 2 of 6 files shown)


__tests__/cmds/versions.test.js
```diff
@@ -37,6 +37,8 @@ jest.mock('../../src/lib/prompts');
 describe('rdme versions*', () => {
   beforeAll(() => nock.disableNetConnect());
 
+  beforeEach(() => jest.resetModules());
```
erunion, line 40: Because we're mocking out `process.env.CIRCLECI` to test the [ci-info](https://npm.im/ci-info) package if we don't reset modules here then that package will get loaded within our `src/lib/enquirer.js` wrapper without the env variable, and cached -- so when the test that does need it, and does mock that environment variable out, the cached module doesn't see it.

__tests__/cmds/versions.test.js
```diff
@@ -103,30 +105,36 @@ describe('rdme versions*', () => {
...
 
-      await expect(createVersion.run({ key, version })).resolves.toBe('Version 1.0.0 created successfully.');
+      await expect(createVersion.run({ key, version, fork: '1.0.0', main: true, beta: false })).resolves.toBe(
+        'Version 1.0.0 created successfully.'
+      );
+
       mockRequest.done();
     });
 
+    it.todo('should prompt a version list if no `--fork` argument is supplied');
```
erunion, line 127: Not having something like `promptHandler.createVersionPrompt.mockResolvedValue` to mock Enquirer responses with this new setup is annoying but as it is I'm comfortable with the tradeoff we're making with this work and not having to have tests for Enquirer being able to prompt us for questions.

src/cmds/versions/create.js
```diff
@@ -37,25 +37,24 @@ module.exports = class CreateVersionCommand {
...
       {
         name: 'beta',
-        type: String,
+        type: Boolean,
         description: 'Is this version in beta?',
       },
       {
-        name: 'isPublic',
-        type: String,
+        name: 'public',
```
erunion, line 49: This is going to be a breaking change for sure, but one I've wanted to do for a while because the `isPublic` argument, it being a string aside, was always out of place with the rest of these.

src/cmds/versions/create.js
```diff
@@ -67,19 +66,45 @@ module.exports = class CreateVersionCommand {
...
 
-    const versionPrompt = promptOpts.createVersionPrompt(versionList || [{}], {
-      newVersion: version,
-      ...opts,
-    });
+    const questions = [
+      {
+        type: 'select',
+        name: 'fork',
+        message: 'Which version would you like to fork from?',
```
erunion, line 81: Having these questions duped in some commands sucks but I'm comfortable with the tradeoff of having duped questions in two commands than having to maintain a question generation library.

src/lib/enquirer.js
```diff
@@ -0,0 +1,58 @@
...
+          }
+        }
+      });
+
+      if (missingArgs.length) {
+        throw new Error(
+          `We've detected you're running within a ${
+            ci.name
+          } environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: ${chalk.yellow(
+            missingArgs.join(',')
```
erunion, line 37: I do **not** like this sentence I've written; please suggest something better.

__tests__/cmds/versions.test.js
```diff
@@ -137,11 +135,32 @@ describe('rdme versions*', () => {
...
-      await expect(createVersion.run({ key, version, fork: '0.0.5' })).rejects.toStrictEqual(
-        new APIError(errorResponse)
-      );
+      await expect(createVersion.run(args)).rejects.toThrow(new APIError(errorResponse));
       mockRequest.done();
     });
+
+    describe('ci environments', () => {
+      beforeEach(() => {
+        process.env.CIRCLECI = true;
```
erunion, line 144: Purposefully testing it with `CIRCLECI` here instead of `GITHUB_ACTIONS` because I know that `GITHUB_ACTIONS` will always be present in our CI and I don't want to delete that from the workflow runner.

__tests__/cmds/versions.test.js
```diff
@@ -103,31 +105,27 @@ describe('rdme versions*', () => {
...
-        from: '1.0.0',
-      });
-
+      const args = { key, version, fork: '1.0.0', main: false, beta: true, public: true };
       const mockRequest = getApiNock()
-        .get('/api/v1/version')
-        .basicAuth({ user: key })
-        .reply(200, [{ version }, { version }])
-        .post('/api/v1/version')
+        .post('/api/v1/version', {
```
erunion, line 110: We weren't fully testing out the expected API payload. Wish there were a way for us to configure Nock to require this...

__tests__/cmds/versions.test.js
kanadgupta: I feel like it might make more sense to do this in a global `beforeEach` call at some point

__tests__/cmds/versions.test.js
kanadgupta: lmaoo yeah I like that we can mock and reuse responses with that `promptHandler`, but we have a lot of different patterns in that file that are pretty confusing so I think we should revisit that at some point

__tests__/cmds/versions.test.js
kanadgupta: good thinking!

__tests__/cmds/versions.test.js
```diff
@@ -137,11 +135,32 @@ describe('rdme versions*', () => {
...
+      });
+
+      it('should prompt for missing required arguments if in a CI environment', async () => {
+        const mockRequest = getApiNock()
+          .get('/api/v1/version')
+          .basicAuth({ user: key })
+          .reply(200, [{ version }, { version }]);
+
+        await expect(createVersion.run({ key, version })).rejects.toThrow(
+          /We've detected you're running within a (.*) environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: fork/
```
kanadgupta, line 158: Any reason why we can't just test for an exact match here?

src/cmds/versions/create.js
kanadgupta: I think having these as booleans makes a lot more sense here, but what does the usage look like? Is it just something like this?
```sh
rdme versions:create 2.0 --key=KEY --public --beta
```

src/cmds/versions/create.js
```diff
@@ -67,19 +66,45 @@ module.exports = class CreateVersionCommand {
...
+        type: 'confirm',
+        name: 'main',
+        message: 'Would you like to make this version the main version for this project?',
+        skip: () => 'main' in opts,
+      },
+      {
+        type: 'confirm',
+        name: 'beta',
+        message: 'Should this version be in beta?',
+        skip: () => 'beta' in opts,
```
kanadgupta, line 95: So just to confirm (since I'm still getting familiar with `enquirer`)... if someone omits the `--beta` flag in a CI environment, `opts.beta` will be populated but set to `false`?

src/cmds/versions/create.js
```diff
@@ -67,19 +66,45 @@ module.exports = class CreateVersionCommand {
...
+        name: 'public',
+        message: 'Would you like to make this version public?',
+        skip() {
+          return 'public' in opts || opts.main || this.enquirer?.answers?.main;
+        },
+      },
+    ];
 
-    const promptResponse = await prompt(versionPrompt);
+    const answers = await enquirer(opts, questions);
```
kanadgupta, line 107: This is probably my only gripe with this approach: could we tweak the patterns defined in `src/lib/enquirer.js` so this matches the usage in the `enquirer` docs? The variable name `enquirer` here is a bit confusing because the `enquirer` object as it's used in the package itself can't be called with the `questions` array like this. Maybe since the default export in `enquirer.js` is  `enquirer.prompt`, we simply rename this default export to prompt?
```suggestion
    const answers = await prompt(opts, questions);
```

src/cmds/versions/create.js
```diff
@@ -89,11 +114,11 @@ module.exports = class CreateVersionCommand {
...
-        codename: codename || '',
-        is_stable: main === 'true' || promptResponse.is_stable,
-        is_beta: beta === 'true' || promptResponse.is_beta,
-        from: fork || promptResponse.from,
-        is_hidden: promptResponse.is_stable ? false : !(isPublic === 'true' || promptResponse.is_hidden),
+        codename: opts.codename || '',
+        is_stable: opts.main || answers.main,
+        is_beta: opts.beta || answers.beta,
+        from: opts.fork || answers.fork,
+        is_hidden: opts.main || answers.main ? false : !('public' in opts || answers.public),
```
kanadgupta, line 121: Nit (and not your fault) but it might be nice to extract these sub-expressions into a separate variables so this is a bit more readable:
```suggestion
        is_hidden: isMain ? false : !isPublic,
```

src/cmds/versions/create.js
kanadgupta: We'll also want to make sure our usage docs are updated: https://github.com/readmeio/rdme#create-a-new-version-1

src/lib/enquirer.js
```diff
@@ -0,0 +1,58 @@
...
+            missingArgs.push(item.name);
+          }
+        }
+      });
+
+      if (missingArgs.length) {
+        throw new Error(
+          `We've detected you're running within a ${
+            ci.name
+          } environment, please supply the following required arguments for your command instead of relying on our terminal prompt system: ${chalk.yellow(
```
kanadgupta, line 36: I'm still recovering from `chalk`'s messiness in the GitHub Actions runner... so as much as I prefer colored text, we should probably remove the `chalk` usage here 😢

src/lib/enquirer.js
```diff
@@ -0,0 +1,58 @@
...
+            missingArgs.join(',')
+          )}`
```
kanadgupta, line 38: This might be too casual (and I'm not sure if prettier will like this), so zhuzh this up as you see fit! I also removed the `chalk` usage:
```suggestion
          `Oops! We noticed you're running this command within a ${
            ci.name
          } environment, but we have a few questions we were going to ask you! We'll need the following arguments for your command to run question-free: ${
            missingArgs.join(',')
          }`
```

src/lib/enquirer.js
```diff
@@ -0,0 +1,58 @@
...
+
+module.exports = function (answers = {}, questions = []) {
+  const enquirer = new Enquirer();
+
+  enquirer.use(disablePromptsInCI(answers));
+
+  return enquirer.prompt(questions).catch(err => {
+    // If the Enquirer promise was rejected without an error then the user killed an open prompt so
+    // we should abide by their wishes and bail out.
+    if (!err) process.exit(0);
```
kanadgupta, line 55: I feel like we could return a `1` here since it's technically not a success, right? Not sure what the best practices are here 🤔

src/lib/enquirer.js
```diff
@@ -0,0 +1,58 @@
...
+    // eslint-disable-next-line no-param-reassign
+    enquirer.prompt = async questions => {
+      const list = [].concat(questions || []);
+
+      // Purposefully loading `ci-info` here so we can adequately test it.
+      const ci = require('ci-info'); // eslint-disable-line global-require
+      if (!ci.isCI) {
+        // If we aren't in a CI environment then we shouldn't look for unanswered questions.
+        return prompt(list);
+      }
```
kanadgupta, line 18: This is really clean 🤩

kanadgupta (changes requested):
I have one general question about the default export from the `enquirer` handler (and a few minor comments/suggestions). Big fan of this pattern overall 👏🏽

src/cmds/versions/create.js
kanadgupta: Ignore me, I asked that before I saw this chunk: https://github.com/readmeio/rdme/blob/f16ae276b5101ac5174f7bc932c94da4e2797060/src/lib/enquirer.js#L22-L30

src/cmds/versions/create.js
erunion: Yeah the usage is that.

__tests__/cmds/versions.test.js
erunion: Because even though I'm setting `CIRCLECI` to the environment, `GITHUB_ACTIONS` is also present and `ci-info` picks that up first so this message says "We've detected you're running within a Github Actions environment" in CI and "within a CircleCI environment" locally.

src/cmds/versions/create.js
erunion: I'm good with that.

src/lib/enquirer.js
erunion: Exit code `1` generally means that that app executed with errors but since in this case I'm killing the prompt myself it's not really an error. Jest also returns the same `0` exit code when you exit out of an open `--watch`.

Exiting Jest before it's able to load any tests:

Screen Shot 2022-02-27 at 7 39 07 PM

Exiting an open `--watch` process:

Screen Shot 2022-02-27 at 7 39 15 PM

src/cmds/versions/create.js
```diff
@@ -71,19 +70,48 @@ module.exports = class CreateVersionCommand {
...
+        type: 'select',
+        name: 'fork',
+        message: 'Which version would you like to fork from?',
+        skip: () => opts.fork,
+        choices: (versionList || [{}]).map(v => ({ message: v.version, value: v.version })),
+      },
+      {
+        type: 'confirm',
+        name: 'main',
+        message: 'Would you like to make this version the main version for this project?',
```
erunion, line 92: @kanadgupta Thoughts on slightly reworking this to the following? The double "version" in there reads off.

```suggestion
        message: 'Would you like to make this the main version for this project?',
```

kanadgupta: Agreed! Nice find

erunion:
After talking through this this morning I'm deciding to close this work out. We have a lot of problems with our existing prompt system (difficult to work with, difficult to test, not every question is mapped to an argument) that while this work aimed to solve it solved it in a backwards way.

Instead of rejecting CI execution of these commands that likely will never be executed from a CI environment anyways, we should instead focus on cleaning up these individual commands instead.

PR #454, opened 2022-02-26
feat(docs): add `--dryRun` option

kanadgupta:
## 🧰 Changes

Realized while working on the docs syncing workflow in #439 that this would be a super easy win and extremely useful to have.
- [x] Adds `--dryRun` option to `docs` command
- [x] Adds `--dryRun` tests
- [x] If someone runs the `docs` command successfully, the output an enormous, mostly unreadable JS array of objects ([example](https://github.com/readmeio/rdme/runs/5340425438?check_suite_focus=true#step:6:19)). Since this tool now supports debugging, this isn't really necessary and isn't particularly user-friendly. So this PR reformats the `docs` command outputs so it always returns formatted success messages doc objects
- [x] Backfilled some test TODOs and updated tests to reflect new command outputs

## 🧬 QA & Testing

If tests pass, we should be good to go! You can also try this out yourself on any project by checking out this branch and running the following command: 
```sh
bin/rdme docs --dryRun --key=API_KEY __tests__/__fixtures__/new-docs
```
This won't make any changes in your project and instead will return a command output that looks like this:
```
🎭 dry run! This will create 'new-doc' with contents from __tests__/__fixtures__/new-docs/new-doc.md with the following metadata: {"category":"5ae122e10fdf4e39bb34db6f","title":"This is the document title"}
```

### DIFF
--- a/src/cmds/docs/index.js
+++ b/src/cmds/docs/index.js
@@ -37,11 +37,16 @@ module.exports = class DocsCommand {
         type: String,
         defaultOption: true,
       },
+      {
+        name: 'dryRun',
+        type: Boolean,
+        description: 'Runs the command without creating/updating any docs in ReadMe. Useful for debugging.',
+      },
     ];
   }
 
   async run(opts) {
-    const { folder, key, version } = opts;
+    const { dryRun, folder, key, version } = opts;
 
     debug(`command: ${this.command}`);
     debug(`opts: ${JSON.stringify(opts)}`);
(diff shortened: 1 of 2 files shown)


src/cmds/docs/index.js
```diff
@@ -83,9 +88,14 @@ module.exports = class DocsCommand {
       return Promise.reject(new Error(`We were unable to locate Markdown files in ${folder}.`));
     }
 
-    function createDoc(slug, file, hash, err) {
+    function createDoc(slug, file, filename, hash, err) {
       if (err.error !== 'DOC_NOTFOUND') return Promise.reject(err);
 
+      if (dryRun)
```
erunion, line 94: Can you wrap add { } to this?

erunion (approved):
Im good to keep this only in v7

PR #463, opened 2022-03-01
fix: a weirdly written promise

erunion:
## 🧰 Changes

Not sure why this promise was written this way but it's weird and I don't like it.


kanadgupta (approved):
lol

PR #479, opened 2022-03-16
docs(GHA): checkout guidance, workflow triggers

kanadgupta:
## 🧰 Changes

Per some feedback from Sarah, I expanded on the event trigger stuff a bit since that might not be super clear to folks. I also updated the docs so we're using `actions/checkout@v3`.

## 🧬 QA & Testing

Do the docs look good?

### DIFF
--- a/documentation/rdme.md
+++ b/documentation/rdme.md
@@ -72,7 +72,7 @@ The command output will indicate whether each page is being created or updated a
 
 ## GitHub Actions Usage
 
-With [GitHub Actions](https://docs.github.com/actions), you can automatically execute workflows when certain events take place in your GitHub repository (e.g. new code is merged into the default branch, a new pull request is opened, etc.).
+With [GitHub Actions](https://docs.github.com/actions), you can automatically execute workflows (e.g. syncing your OpenAPI definition or Markdown docs to ReadMe) when certain events take place in your GitHub repository (e.g. code is pushed to the default branch, a new pull request is opened, etc.). There are literally [dozens of event options available](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows), making the possibilities endless!
 
 > 📘 Keeping `rdme` up-to-date
 >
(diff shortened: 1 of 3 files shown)


erunion (approved):

documentation/rdme.md
```diff
@@ -72,7 +72,7 @@ The command output will indicate whether each page is being created or updated a
 
 ## GitHub Actions Usage
 
-With [GitHub Actions](https://docs.github.com/actions), you can automatically execute workflows when certain events take place in your GitHub repository (e.g. new code is merged into the default branch, a new pull request is opened, etc.).
+With [GitHub Actions](https://docs.github.com/actions), you can automatically execute workflows (e.g. syncing your OpenAPI definition or Markdown docs to ReadMe) when certain events take place in your GitHub repository (e.g. code is pushed to the default branch, a new pull request is opened, etc.). There are literally [dozens of event options available](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows), making the possibilities endless!
```
RyanGWU82, line 75: Although this is a good description of Actions generally, it might be more helpful if we provided some guidance that's more specific to `rdme`. "There are dozens of event options available, but you'll typically want to sync your OpenAPI definition and Markdown docs to ReadMe [when a new build is deployed](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#deployment) or [when a new release is created](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release). If you're not using either of those features then you can sync to ReadMe [when code is pushed to the default branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push)."

RyanGWU82 (approved):

documentation/rdme.md
kanadgupta: 1000% agreed, thank you for the feedback! Will incorporate!

PR #486, opened 2022-03-31
feat: throwing error messaging on node 12 being unsupported

erunion:
## 🧰 Changes

We've been getting a healthy amount of support questions about rdme throwing an "Unexpected token '.'" error and because this error is thrown on Node versions that don't support [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) I'm adding a hard top into `bin/rdme` if it's run with Node 12 or below.

## 🧬 QA & Testing

There's a new test here for this that should only be run on Node 12. I've tested it out locally and it works, hoping it still works with this forked job setup.


erunion:
Confirmed that the Node 12 test and passed in CI:

Screen Shot 2022-03-31 at 3 23 20 PM

bin/rdme
```diff
@@ -20,7 +32,7 @@ require('../src')(process.argv.slice(2))
       'support@readme.io'
     )}.`;
 
-    if (err?.message) {
+    if (err.message) {
```
erunion, line 35: Removed optional chaining from here because Node 12 wouldn't run this file at all with it (even though the `require(../src)` also loads code with it.

bin/rdme
```diff
@@ -2,13 +2,25 @@
...
 const isGHA = require('../src/lib/isGitHub');
 
 updateNotifier({ pkg }).notify();
 
+/**
+ * We use optional chaining throughout the library, which doesn't work on Node 12, so to curb
+ * support questions about why rdme is throwing an "Unexpected token '.'" error we should hard
+ * stop if we're on any Node version that doesn't support it.
+ */
+if (nodeVersion.major <= 12) {
```
kanadgupta, line 18: Since we already have `semver` in the mix, could we use the `engines.node` field and have `semver` check if the current version satisfies the requirements?

__tests__/bin.test.js
```diff
@@ -0,0 +1,28 @@
+const { exec } = require('child_process');
```
kanadgupta, line 1: ugh thank you for adding tests for this file

__tests__/bin.test.js
```diff
@@ -0,0 +1,28 @@
...
+const nodeVersion = require('parse-node-version')(process.version);
+
+describe('bin', () => {
+  if (nodeVersion.major > 12) {
```
kanadgupta, line 5: this also feels like another place we could leverage our `engines.node` field for, but I'm not as concerned about this.

bin/rdme
```diff
@@ -2,13 +2,25 @@
...
+  const message = `We're sorry, this release of rdme does not support your Node version of ${nodeVersion.major}.`;
```
kanadgupta, line 19: This feels like a good place to just toss in the `engines.node` value:
```suggestion
  const message = `We're sorry, this release of rdme does not support Node.js v${nodeVersion.major}. We support the following versions: ${node}`;
```

__tests__/bin.test.js
```diff
@@ -0,0 +1,28 @@
...
+        });
+      });
+    });
+  } else {
+    it('should fail with a message', async () => {
+      expect.assertions(1);
+
+      await new Promise(done => {
+        exec(`node ${__dirname}/../bin/rdme`, (error, stdout, stderr) => {
+          expect(stderr).toContain("We're sorry, this release of rdme does not support your Node version of 12.");
```
kanadgupta, line 22:
```suggestion
          expect(stderr).toContain("We're sorry, this release of rdme does not support Node.js v12.");
```

kanadgupta (changes requested):
Overall approach + tests LGTM, but I think we should use `semver` + our `engines.node` field instead of hardcoding our Node.js restrictions wherever possible.

kanadgupta (approved):

PR #489, opened 2022-04-01
fix: messy issues with nodeVersionUtils helper needing semver installed

erunion:
## 🧰 Changes

I broke our doc syncing workflow by moving `getNodeVersion()` into a new helper with `isSupportedNodeVersion()`, which requires `semver` installed. We don't need or want to run a `npm ci` in our doc syncing workflow and its broken without this right now so I'm reverting some of my reorg in #486.


erunion:
Screen Shot 2022-04-01 at 9 56 34 AM

🏌️

kanadgupta (approved):

PR #493, opened 2022-04-14
feat(help): typos, add `rdme` version

kanadgupta:
## 🧰 Changes

lol... just discovered we spell "utility" wrong in the main help screen 😮‍💨 

```
$ rdme

                  📖 rdme

    a utlity for interacting with ReadMe
```
This fixes that, and also adds the current `rdme` version to the help screen:
```
Options

  -h, --help       Display this usage guide               
  -v, --version    Show the current rdme version (v7.1.1) 
```

## 🧬 QA & Testing

Unfortunately this means I had to rip out a snapshot test in favor of a partial test, but if tests pass and the copy looks good, we should be good to go!


kellyjosephprice (approved):
`utlity` is disrupting municipal payment systmes

__tests__/bin.test.js
```diff
@@ -9,7 +9,7 @@ describe('bin', () => {
 
       await new Promise(done => {
         exec(`node ${__dirname}/../bin/rdme`, (error, stdout) => {
-          expect(stdout).toContain('a utlity for interacting with ReadMe');
+          expect(stdout).toContain('a utility for interacting with ReadMe');
```
erunion, line 12: couldn't uodate the snapshot?

erunion (approved):

__tests__/bin.test.js
kanadgupta: problem is that the snapshot contains the package version, and we'd have to update the snapshot for every bump and I couldn't think of an easy way to get around that

PR #521, opened 2022-06-23
ci(CodeQL): add config

kanadgupta:
## 🧰 Changes

This adds a config file [per GitHub's docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan) so CodeQL ignores our test files.

## 🧬 QA & Testing

Does CodeQL pass and do the changes look sound?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,4 +1,4 @@
-name: "CodeQL"
+name: 'CodeQL'
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -17,19 +17,15 @@ jobs:
...
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v2
         with:
-          languages: ${{ matrix.language }}
+          config-file: ./.github/codeql/config.yml
+          languages: javascript
```
kanadgupta, line 28: Per [their docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#changing-the-languages-that-are-analyzed) this can be simplified so I just removed the matrix.

PR #530, opened 2022-06-29
feat: add `categories` and `categories:create`

garrett-wade:
| 🚥 Fix RM-XXX |
| :-- |

## 🧰 Changes

Added support for two new commands `categories` and `categories:create`. 
- `categories` returns a raw json of all categories for a version including pagination.
- `categories:create` allows you to create a new category with a specified `title` and `type` for your project version. The `title` will be converted to a slug in param case. If a category already exists with the same slug and type for
the specified version then a new category is not created and the id of the existing category is returned. This includes if the slug contains incremented suffix such as `-1` or `-2`. If there are multiple categories with a matching slug and type then only one matching category is returned.

## 🧬 QA & Testing

- `categories`
    1. Login to a ReadMe project with `rdme login`
    2. Run `rdme categories`
- `categories:create`
   1. Login to readme project with `rdme login`
   2. Run `rdme categories:create "test title" --categoryType guide` and see a new category was created with a slug of `test-title`
   3. Run `rdme categories:create "test title" --categoryType guide` and see a new category was not created but rather it returns the slug and id of the category created in the previous step.
   4. Run `rdme categories:create "test title" --categoryType reference` and see a new category was created with a slug of `test-title-1`
   5. Run `rdme categories:create "test title" --categoryType reference` and see a new category was not created but rather it returns the slug and id of the category created in the previous step.
   6.  Run `rdme categories:create "test title"` and see a `categoryType` of "guide" or "reference" is required


package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "@actions/core": "^1.6.0",
     "chalk": "^4.1.2",
+    "change-case": "^4.1.2",
```
erunion, line 39: Can you include the `package-lock.json` changes for this? The test suite is failing right now because it's missing.

erunion (changes requested):

src/cmds/categories/index.js
```diff
@@ -0,0 +1,72 @@
...
+        }),
+      }).then(res => {
+        return Math.ceil(res.headers.get('x-total-count') / 10);
+      });
+    }
+
+    const allCategories = [].concat(
+      ...(await Promise.all(
+        Array.from({ length: await getNumberOfPages() }, (_, i) => i + 1).map(async page => {
+          return fetch(`${config.get('host')}/api/v1/categories?perPage=10&page=${page}`, {
```
erunion, line 59:
```suggestion
          return fetch(`${config.get('host')}/api/v1/categories?perPage=20&page=${page}`, {
```

src/lib/commands.js
```diff
@@ -88,5 +88,9 @@ exports.getCategories = () => {
       description: 'Other useful commands',
       commands: [],
     },
+    categories: {
```
erunion, line 91: Can you move this above `utilities` so it these category commands won't be at the very end of the `--help` screen?

src/cmds/categories/index.js
```diff
@@ -0,0 +1,72 @@
...
+const { getProjectVersion } = require('../../lib/versionSelect');
+const fetch = require('../../lib/fetch');
+const { cleanHeaders, handleRes } = require('../../lib/fetch');
+const { debug } = require('../../lib/logger');
+
+module.exports = class CategoriesCommand {
+  constructor() {
+    this.command = 'categories';
+    this.usage = 'categories [options]';
+    this.description = 'Get all catefories in your ReadMe project';
```
erunion, line 11:
```suggestion
    this.description = 'Get all categories in your ReadMe project';
```

erunion (changes requested):

garrett-wade:
@erunion updates complete

src/cmds/categories/create.js
```diff
@@ -0,0 +1,121 @@
...
+    async function matchCategory() {
+      return allCategories.filter(category => {
+        return category.slug.match(`${slug}(-\\d)?`) && category.type === categoryType;
+      });
+    }
+
+    async function createCategory() {
+      const matchedCategory = await matchCategory();
+      if (matchedCategory.length > 0) {
+        return `The '${matchedCategory[0].slug}' category with a type of '${matchedCategory[0].type}' already exists with an id of '${matchedCategory[0].id}'. A new category was not created`;
```
kanadgupta, line 100: I'm still going back and forth on if this slug uniqueness check is necessary. The category slug isn't surfaced in the docs, it's more of a behind-the-scenes identifier.

Let's say a project contains the category slug `example`. If a user attempts to create the category "Example", the category will be successfully created with a slug of `example-1`. I think as long as we report back what the slug is in the success response, this behavior should be fine—right?

src/cmds/categories/create.js
```diff
@@ -0,0 +1,121 @@
...
+        }),
+      }).then(res => {
+        return Math.ceil(res.headers.get('x-total-count') / 20);
+      });
+    }
+
+    const allCategories = [].concat(
+      ...(await Promise.all(
+        Array.from({ length: await getNumberOfPages() }, (_, i) => i + 1).map(async page => {
+          return fetch(`${config.get('host')}/api/v1/categories?perPage=20&page=${page}`, {
```
kanadgupta, line 80: If we do keep this slug check logic intact, I think we should look for the slug via [this endpoint](https://docs.readme.com/reference/getcategory) instead of fetching every category and iterating through all of them.

kanadgupta (changes requested):
Thanks for the PR @garrett-wade! I have a few overarching questions about approach on the `create` command below. Would love your feedback as a user of this tool!

src/cmds/categories/create.js
garrett-wade: Hey @kanadgupta, 

Our development team is publishing documentation to ReadMe via API and are leveraging the `docs` command to publish our guides. Our engineers need to easily create a new or retrieve the existing unique category id for their feature to add to their markdown file's grey-matter. The category title maps to the feature name in our product taxonomy, and we only want one guide category and one reference category per feature. I chose the slug to identify duplicates as it had a more consistent format than title, but suppose the match could be done on the title as well. Curious to hear your thoughts on this.

src/cmds/categories/create.js
garrett-wade: @kanadgupta I thought about this, but ran into some issues. For Example:

1. Developer 1 uses the `categories:create` command to create a category with a title of `Test` and a type of `reference`. 
2. This would create a category with a slug of `test` and a type of `reference`
3. Developer 1 uses the `categories:create` command to  create a category with a title of `Test` and a type of `guide`
4. This would create a category with a slug of `test-1` and a type of `guide`
5. Developer 2 uses the `categories:create` command to  create a category with a title of `Test` and a type of `guide`
6. This would use [this endpoint](https://docs.readme.com/reference/getcategory) to search for an existing category with a slug of `test`, but because the category with a type of `guide`  in step 3 has a slug of `test-1` than a category will not be returned and a duplicate category of `Test` and a type of `guide` will be created with a slug of `test-2`.

Because the slug has an automatically incremented suffix, the user using the command will have to know that the suffix exists to search properly and to do that they will need to call [this endpoint](https://docs.readme.com/reference/getcategories) to determine that anyway.

kanadgupta (changes requested):
Thanks for the context @garrett-wade! I think the example makes sense and I can see title-uniqueness checks being useful. Could you make the following changes:

- [ ] Could you refactor your partial-slug matching to do exact title-matching instead? Partial slug matching may incur some weird false positives and title-matching seems more robust.

- [ ] Could you wrap the aforementioned matching logic in a flag called `preventDuplicateTitles`? In general I think the default behavior should follow the behavior of our API and the dashboard, but this way users can opt-in to ensuring that titles are unique.

- [ ] Can you merge the `main` branch into this branch and re-run `npm install` with at least `npm@7`? It appears [tests are still failing](https://github.com/readmeio/rdme/runs/7219167057?check_suite_focus=true).

src/cmds/categories/create.js
```diff
@@ -0,0 +1,127 @@
...
+              'x-readme-version': selectedVersion,
+              Accept: 'application/json',
+            }),
+          }).then(res => handleRes(res));
+        })
+      ))
+    );
+
+    async function matchCategory() {
+      return allCategories.filter(category => {
```
kanadgupta, line 96: Small performance suggestion: switching to [`Array.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here will have it stop iterating once it's found what it's looking for (plus, it returns a boolean!):

```suggestion
      return allCategories.some(category => {
```

src/cmds/categories/create.js
```diff
@@ -0,0 +1,127 @@
...
+        return changeCase.paramCase(category.title) === changeCase.paramCase(title) && category.type === categoryType;
```
kanadgupta, line 97: Do we still need `changeCase` here?
```suggestion
        return category.title === title && category.type === categoryType;
```

kanadgupta (changes requested):
Thanks for making those suggested changes @garrett-wade! Some more comments below:

src/cmds/categories/create.js
```diff
@@ -0,0 +1,127 @@
...
+      });
+    }
+
+    async function createCategory() {
+      if (preventDuplicates) {
+        const matchedCategory = await matchCategory();
+        if (matchedCategory.length > 0) {
+          return `The '${matchedCategory[0].title}' category with a type of '${matchedCategory[0].type}' already exists with an id of '${matchedCategory[0].id}'. A new category was not created`;
```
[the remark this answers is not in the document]
garrett-wade: @kanadgupta I am not sure this is truly an error condition, as this may allow the end user to actually "check for duplicates" and that may be the actual intention.

src/cmds/categories/create.js
garrett-wade: @kanadgupta Agree from a performance standpoint, but we actually want to return the category object as some of it's attributes such as the `id` are used in the response when a duplicate is found.

src/cmds/categories/create.js
garrett-wade: @kanadgupta I used the changeCase library here to ensure matches on capitalization and spaces. As to ensure an exact match. Open to other methods here.

src/cmds/categories/create.js
kanadgupta: Ah my apologies, I missed the part where we use the found object's attributes. Could you switch to [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead? That way we get the aforementioned performance benefits and can use the object properly.

```suggestion
      return allCategories.find(category => {
```

src/cmds/categories/create.js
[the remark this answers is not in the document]
kanadgupta: Yeah per @erunion's comment, the intent of this command is to create a category, therefore we should return an error if it fails to create a category.

src/cmds/categories/create.js
erunion: I'm comfortable just doing a `.toLowerCase()` check on these instead of pulling `change-case` in for this one line.

src/cmds/categories/create.js
```diff
@@ -0,0 +1,127 @@
...
+    );
+
+    async function matchCategory() {
+      return allCategories.filter(category => {
+        return changeCase.paramCase(category.title) === changeCase.paramCase(title) && category.type === categoryType;
+      });
+    }
+
+    async function createCategory() {
+      if (preventDuplicates) {
```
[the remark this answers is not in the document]
kanadgupta: Are you still working on this? This comment got marked as resolved but it's still applicable so I'm going to unresolve it! Just to reiterate, the `getNumberOfPages ` and `matchCategory` functions should not be running at all if `preventDuplicates` is set to false.

kanadgupta (changes requested):

src/cmds/categories/create.js
kanadgupta: Also if you could add a test case (i.e. with "Example Category" and "Example category "), that'd be great!

garrett-wade:
@kanadgupta updated based on requests above, for `preventDuplicates` I do no think either of the above methods are ran if the flag is checked with the current implementation.

garrett-wade:
@erunion @kanadgupta Thanks for all the collaboration and feedback on this. Learned a lot and enjoy working with both of you!

PR #538, opened 2022-07-20
feat: better error-handling for non-JSON responses

kanadgupta:
| 🚥 Fix RM-4868 |
| :-- |

## 🧰 Changes

Right now, if we receive an error response from the API when running the `openapi` command, we have some not-so-great error handling where if the response body isn't JSON, we basically just write it off as a timeout error: https://github.com/readmeio/rdme/blob/862d4009257ddbbc553cbe67fe8a4ebb1b7bec7c/src/cmds/openapi.js#L111-L129

Our debug logs in this situation aren't helpful either: since we blindly use `res.json()` to parse the response body, our debug logs consistently return an unhelpful `Unexpected token < in JSON` error.

In this PR, our approach to parsing the response from our API has been improved to better allow us to understand non-JSON response bodies. We now do some smarter inspection of the response to determine whether or not it's a timeout. It probably is in most cases, but now we can actually inspect the response body and determine that.

In order to do this, I refactored our generic `handleRes` handler to dynamically handle JSON and non-JSON responses better. I was able to eliminate some duplicate code and the `openapi` command now leverages uses `handleRes` for error-handling.

## 🧬 QA & Testing

Do tests pass and do the changes look good?

If you want to try this out with an OAS file that timeouts consistently, check out [this comment](https://linear.app/readme-io/issue/RM-4464#comment-21c5b027). I confirmed locally with that file that the timeout error is thrown properly.

Screen Shot 2022-07-20 at 5 54 49 PM

I also confirmed that the debug logs now properly log non-JSON response bodies properly:

image

### DIFF
--- a/__tests__/cmds/openapi.test.js
+++ b/__tests__/cmds/openapi.test.js
@@ -535,7 +535,11 @@ describe('rdme openapi', () => {
 
       await expect(
         openapi.run({ spec: require.resolve('@readme/oas-examples/2.0/json/petstore.json'), key, version })
-      ).rejects.toStrictEqual(new Error('There was an error uploading!'));
+      ).rejects.toStrictEqual(
+        new Error(
+          'Yikes, something went wrong! Please try uploading your spec again and if the problem persists, get in touch with our support team at support@readme.io.'
+        )
+      );
 
       return mock.done();
     });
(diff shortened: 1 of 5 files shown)


__tests__/cmds/openapi.test.js
```diff
@@ -555,7 +559,7 @@ describe('rdme openapi', () => {
         .post('/api/v1/api-specification', { registryUUID })
         .delayConnection(1000)
         .basicAuth({ user: key })
-        .reply(400, '<html></html>');
+        .reply(400, '<title>Application Error</title>');
```
kanadgupta, line 562: Before, we just assumed any ol' HTML was a timeout. Yikes! Now, we actually look for snippets of the Heroku/Cloudflare error pages.

erunion, line 562:
```suggestion
        .reply(500, '<title>Application Error</title>');
```

erunion (approved):
awesome

__tests__/cmds/openapi.test.js
kanadgupta: oh damn great catch thank you!

PR #540, opened 2022-07-21
refactor(docs): consolidate logic

kanadgupta:
## 🧰 Changes

Follow-up to #529. Consolidates our docs create/update logic into a single function. This should set up future functionality (e.g. Custom Pages, Changelog) more easily.

## 🧬 QA & Testing

Tests haven't changed at all, so if those pass then we should be good to go!


erunion (approved):
DRY humidity > wet

PR #542, opened 2022-07-25
refactor: split out fixtures, test cleanup

kanadgupta:
## 🧰 Changes

Just some refactoring as part of the upcoming Changelog + Custom Pages functionality.

- [x] Refactoring our generic `pushDoc` function to support more routes and check for generic 404s as opposed to `DOC_NOTFOUND` errors
- [x] Some minor grammar + camelcasing changes (i.e. `filePath` instead of `filepath`)
- [x] Moving all the fixtures for the `docs` command into its own folder

## 🧬 QA & Testing

Do tests pass?


src/cmds/docs/single.js
```diff
@@ -8,12 +8,12 @@ const pushDoc = require('../../lib/pushDoc');
 module.exports = class SingleDocCommand {
   constructor() {
     this.command = 'docs:single';
-    this.usage = 'docs:single <filepath> [options]';
+    this.usage = 'docs:single <file> [options]';
```
erunion, line 11: Doesn't `<file>` need to match the arg that we're going to be using (`filePath` in this case)?

src/cmds/docs/single.js
```diff
@@ -62,7 +62,7 @@ module.exports = class SingleDocCommand {
 
     debug(`selectedVersion: ${selectedVersion}`);
 
-    const createdDoc = await pushDoc(key, selectedVersion, dryRun, filepath);
+    const createdDoc = await pushDoc(key, selectedVersion, dryRun, filePath, this.category);
```
erunion, line 65: Where's `this.category` being set?

src/cmds/docs/single.js
kanadgupta: Nope, this is just the string used for the usage example

erunion: It's been so long since I've looked at this part of the code but how does it know to give you the first arg as `filePath`?

src/cmds/docs/single.js
kanadgupta: We pass that into the `pushDoc` function so we can use it for dynamic routing:

https://github.com/readmeio/rdme/blob/aaad56d984da47ed3576a8b672de5b848a07ff8d/src/lib/pushDoc.js#L42

The idea is we can use a single handler for `docs`, `changelogs`, and `custompages` routes

src/cmds/docs/single.js
erunion: I swear I thought it was parsing the usage string to determine if things were required or not

kanadgupta: this guy: https://github.com/readmeio/rdme/blob/aaad56d984da47ed3576a8b672de5b848a07ff8d/src/cmds/docs/single.js#L28-L32

src/cmds/docs/single.js
erunion: OH right, `this.category` is the category that it shows up in the help screen.

src/cmds/docs/single.js
erunion: Ah there it is. Ignore me today!

erunion (approved):

src/cmds/docs/single.js
kanadgupta: lol misread your question so my response didn't actually answer your question but yes, it's for the categories in the help screen

PR #545, opened 2022-07-28
feat: add retag-release workflow to add a vX.Y.Z tag for our GitHub Action

RyanGWU82:
| 🚥 Fix RM-4840 |
| :-- |

## 🧰 Changes

The **rdme** GitHub Action required people to write workflows with the line `uses: readmeio/rdme@1.2.3` but GitHub's convention is for action versions to be tagged `v1.2.3`. Because of this some users have written `uses: readme/rdme@v1.2.3` and of course they expect that to work.

This workflow looks for releases created with a tag like `1.2.3` and adds a second tag like `v1.2.3`. It uses [actions/github-script](https://github.com/actions/github-script) which is the simplest way to run a short script in a GitHub workflow. There were some existing third-party actions for adding a tag to a repo, but I didn't really want to rely on a third-party script for something so simple. (Admittedly this was going to be a one-liner but with input validation and HTTP error checking it got a bit longer.)

## 🧬 QA & Testing

Ran this on a little test project and it worked as expected:

Screen Shot 2022-07-28 at 11 15 10 AM

Exits early when the tag doesn't follow semantic versioning style, so that we don't create tags like `vv1.2.3` or `vdeprecated`.

Screen Shot 2022-07-28 at 11 15 01 AM

### DIFF
--- a/.github/workflows/retag-release.yaml
+++ b/.github/workflows/retag-release.yaml
@@ -1,0 +1,53 @@
+# This script adds a "vX.Y.Z" tag to every new release.
+#
+# Our releases are tagged like "1.2.3" but we want people to be able to write
+# GitHub Action workflows to say "uses: readmeio/readme@v1.2.3" because that's
+# the usual GitHub convention.
+
+name: retag-release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  retag-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6.1.0
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const oldTag = context.payload.release.tag_name;
+            if (!oldTag.match(/^[0-9]+\.[0-9]+\.[0-9]+$/)) {
+              console.log('Not retagging this release: This script will only retag releases that use');
+              console.log(`semantic versioning, like "1.2.3", but this release's tag is "${oldTag}".`);
+              return;
+            }
+            const newTag = `v${oldTag}`;
+            console.log(`Retagging release "${oldTag}" as "${newTag}".`);
+
+            const oldRef = await github.rest.git.getRef({
+              owner,
+              repo,
+              ref: `tags/${oldTag}`,
+            });
+            if (oldRef.status < 200 || oldRef.status >= 400) {
+              console.log(oldRef);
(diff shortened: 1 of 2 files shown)


.github/workflows/retag-release.yaml
```diff
@@ -0,0 +1,53 @@
...
+            });
+            if (oldRef.status < 200 || oldRef.status >= 400) {
+              console.log(oldRef);
+              throw new Error(`GitHub API call returned HTTP status code ${oldRef.status}`);
+            }
+            const sha = oldRef.data.object.sha;
+            console.log(`Found tag "${oldTag}"; commit hash is ${sha}`);
+
+            console.log(`Creating tag "${newTag}" pointing to commit hash ${sha}...`);
+            const newRef = await github.rest.git.createRef({
```
erunion, line 43: Will this new tag/release get created with the same release notes?

RyanGWU82: It's only a tag, not a release, so it won't have any release notes. (It will show up as a distinct tag on https://github.com/readmeio/rdme/tags but not as a separate release on https://github.com/readmeio/rdme/releases.)

erunion (approved):
seems good to me!

.github/workflows/retag-release.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  retag-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6.1.0
```
kanadgupta, line 17:
```suggestion
      - uses: actions/github-script@v6
```

.github/workflows/retag-release.yaml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          script: |
+            const { owner, repo } = context.repo;
```
kanadgupta, line 20: Thoughts on popping this script into a separate JS file? https://github.com/actions/github-script#run-a-separate-file

That way we can lint it, run it locally more easily, etc.

RyanGWU82: Oh cool, didn't notice that. The syntax for doing that is a little janky but I like that idea a lot more than embedding it here in the YAML. New commit coming up shortly…

PR #547, opened 2022-07-28
feat: custom pages support

kanadgupta:
| 🚥 Fix RM-4889 |
| :-- |

## 🧰 Changes

Adds support for the [Custom Pages endpoints](https://docs.readme.com/reference/getcustompages) to `rdme`. Will be updating [`documentation/rdme.md`](https://github.com/readmeio/rdme/blob/main/documentation/rdme.md) in a later PR.

## 🧬 QA & Testing

The test suite should look familiar, but only real differences to call out is that we support HTML files in Custom Page uploads, so we include a handful of tests for that.


README.md
```diff
@@ -191,6 +191,26 @@ This command also has a dry run mode, which can be useful for initial setup and
 rdme changelogs:single path-to-markdown-file
 ```
 
+### Custom Pages
+
+#### Syncing a Folder of Custom Pages to ReadMe
```
erunion, line 196: God, I wish these docs were pulled directly from the command files. `oclif` has something like this. Would be neat.

__tests__/cmds/custompages.test.js
```diff
@@ -0,0 +1,710 @@
...
+    let simpleDoc;
+    let anotherDoc;
+
+    beforeEach(() => {
+      let fileContents = fs.readFileSync(path.join(fullFixturesDir, '/existing-docs/simple-doc.md'));
+      simpleDoc = {
+        slug: 'simple-doc',
+        doc: frontMatter(fileContents),
+        hash: hashFileContents(fileContents),
+      };
```
erunion, line 64: Super nitpicky but might be cleaner to do these as a promise?

```suggestion
simpleDoc = await fsPromises.readFile(path.join(fullFixturesDir, '/existing-docs/simple-doc.md')).then(contents => ({
  slug: 'simple-doc',
  doc: frontMatter(contents),
  hash: hashFileContents(contents),
});
```

src/lib/pushDoc.js
```diff
@@ -32,6 +32,16 @@ module.exports = async function pushDoc(key, selectedVersion, dryRun, filepath,
   const slug = matter.data.slug || path.basename(filepath).replace(path.extname(filepath), '').toLowerCase();
   const hash = crypto.createHash('sha1').update(file).digest('hex');
 
+  let data = { body: matter.content, ...matter.data, lastUpdatedHash: hash };
+
+  if (type === 'custompages') {
+    if (filepath.endsWith('.html')) {
```
erunion, line 38: Related to my comment in that other PR but we might want to drop case-sensitive checking on these extensions so we can allow `.HTML` through.

erunion (approved):

src/lib/pushDoc.js
kanadgupta: Fixed this here: https://github.com/readmeio/rdme/pull/549/commits/706cdaf8ec5011d16164a51f8406bf4a7a20c53f

README.md
kanadgupta: Agreed 😭

__tests__/cmds/custompages.test.js
kanadgupta: feels like two steps forward two steps back so gonna hold off on this!

erunion: lol

PR #549, opened 2022-07-28
fix(docs/single): support files with `.markdown` extension

### LINKED ISSUE #546: Files ending in `.markdown` don't sync properly
These lines are wrong:
https://github.com/readmeio/rdme/blob/fe2498415e3a8ff46646a90876403513a4135259/src/cmds/docs/single.js#L54
https://github.com/readmeio/rdme/blob/fe2498415e3a8ff46646a90876403513a4135259/src/cmds/changelogs/single.js#L48

We should fix this and add a test to check for these files.

kanadgupta:
| 🚥 Fix #546  |
| :-- |

## 🧰 Changes

Fixes an issue where our single commands had faulty conditional logic and therefore weren't working on `.markdown` files properly.

Also leans up a few tests, see https://github.com/readmeio/rdme/commit/21ba907ca774230366ff2fee6d872babe02137ef

## 🧬 QA & Testing

Do the test changes look good and do they pass?

### DIFF
--- a/src/cmds/changelogs/single.js
+++ b/src/cmds/changelogs/single.js
@@ -45,7 +45,7 @@ module.exports = class SingleChangelogCommand {
       return Promise.reject(new Error(`No file path provided. Usage \`${config.get('cli')} ${this.usage}\`.`));
     }
 
-    if (!(filePath.endsWith('.md') || filePath.endsWith('.markdown'))) {
+    if (!(filePath.toLowerCase().endsWith('.md') || filePath.toLowerCase().endsWith('.markdown'))) {
       return Promise.reject(new Error('The file path specified is not a Markdown file.'));
     }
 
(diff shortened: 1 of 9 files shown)


src/cmds/changelogs/single.js
```diff
@@ -45,7 +45,7 @@ module.exports = class SingleChangelogCommand {
       return Promise.reject(new Error(`No file path provided. Usage \`${config.get('cli')} ${this.usage}\`.`));
     }
 
-    if (filePath.endsWith('.md') === false || !filePath.endsWith('.markdown') === false) {
+    if (!(filePath.endsWith('.md') || filePath.endsWith('.markdown'))) {
```
erunion, line 48: I've had case-sensitive filesystems on my mind this week and I'm wondering if we should support `.MD` and `.Markdown`, or any casing variation of them.

src/cmds/changelogs/single.js
```diff
@@ -45,7 +45,7 @@ module.exports = class SingleChangelogCommand {
...
       return Promise.reject(new Error('The file path specified is not a markdown file.'));
```
erunion, line 49: Does this need to be capitalized?

```suggestion
      return Promise.reject(new Error('The file path specified is not a Markdown file.'));
```

erunion (approved):

PR #557, opened 2022-08-01
feat(logger): add GHA-aware notices/warnings

kanadgupta:
| 🚥 Fix RM-4986 |
| :-- |

## 🧰 Changes

Added some context-aware logging for [notices and warnings](https://github.com/actions/toolkit/tree/main/packages/core#annotations) (we were previously only doing this for if `rdme` would error out). This will surface certain logs for the user more nicely when it's being run in a GitHub Actions environment. We'll need this for the work in RM-4984.

I also added [`ci-detect`](https://github.com/npm/ci-detect) into the mix since we will also need that for RM-4984 and it feels a little less hairy to do our CI detection checks this way.

## 🧬 QA & Testing

I tested out the annotations and confirmed that they show up! You can check it out here: https://github.com/readmeio/rdme/actions/runs/2778155329


__tests__/lib/fetch.test.js
```diff
@@ -9,6 +9,7 @@ describe('#fetch()', () => {
     // List of all GitHub Actions env variables:
     // https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
     beforeEach(() => {
+      process.env.GITHUB_ACTION = '__repo-owner_name-of-action-repo';
```
kanadgupta, line 12: I added this to further simulate a GHA environment in our tests since this is the variable that [`ci-detect`](https://github.com/npm/ci-detect) uses.

erunion (approved):

PR #558, opened 2022-08-03
feat: refactoring the codebase to be ESM

erunion:
## 🧰 Changes

Our `update-notifier` dependency currently has a CVE for its nested `got` dependency. Unfortunately `update-notifier` has moved to being ESM-only in v6 and we can't use it without moving this codebase over to ESM.

That's this.

## 🐳 Problems

* [ ] Most of the test suite currently fails because `__dirname` is unavailable with in ES scope in Node. I've polyflled it in a few spots but I'm not sure if thats the right solution or if we could replace it with a `process.cwd()` call instead.
* [ ] Jest `--watch` is currently broken because of the way that we're loading commands for the help screen. Jest will run tests the first time just fine but if you make any changes it'll fail with a "module is not a module" error because we're dynamically loading in commands with `await import()` and Jest has already town down its environment. Instead of futzing with this it might be best to move away from dynamic command loading and have an `index` file in the `cmds/` directory that exports everything.
* [ ] I'm still unclear if having `type: module` in `rdme` will force you to have `type: module` in any package that has this as a dependency, or if it's fine because nobody is doing `require('rdme')`.
* [ ] Jest's ESM handling does not like `chalk`. At all. I've had to do some module resolution in the config to get it working and it's unfortunately causing us to not be able to upgrade chalk to its latest ESM-only version.


erunion:
Alternatively we could say nuts to any package that goes ESM-only and replace it with something else.

* `update-notifier` could be replaced with some code like NPM has in https://github.com/npm/cli/blob/latest/lib/utils/update-notifier.js
* `chalk` could be replaced with `ansi-colors`.
* We can/should replace `node-fetch` with an `api`-generated SDK.

erunion:
Closing this out for now in favor of https://github.com/readmeio/rdme/pull/560.

PR #559, opened 2022-08-08
feat(openapi): version handling improvements

kanadgupta:
| 🚥 Fix RM-4898 RM-4984 |
| :--------------------- |

## 🧰 Changes

- [x] Added CI auto-detection to our version selection prompt (used in `openapi`, `docs`, and `categories` commands). When the user fails to pass a `--version` parameter in a CI environment, we just default to the main version (and inform the user of this via a warning—see below)
  image
- [x] Added a new `--useSpecVersion` parameter to the `openapi` command that uses the version in the `info.version` field of the spec for the ReadMe project version
- [x] Consolidated `--version` parameter description into a separate file called `src/lib/versionOpt.js` where it was appropriate to do so
- [x] Small tweaks to a few parameter descriptions
- [x] Added a li'l ℹ️ emoji prefix to our `info` statements

## 🧬 QA & Testing

I added test coverage for all this new version-handling logic. Alternatively, you can checkout this branch and run the following command to confirm that the command will try to sync to `v1.2.3` of your project:

```sh
bin/rdme openapi --useSpecVersion __tests__/__fixtures__/petstore-simple-weird-version.json
```

image


kanadgupta:
@erunion @Dashron this has the TS changes merged in and is ready for review!

__tests__/lib/commands.test.ts
```diff
@@ -56,7 +56,9 @@ describe('utils', () => {
                   }
 
                   expect(arg.description).toBe(
-                    command.command !== 'versions' ? 'Project version' : 'A specific project version to view'
+                    command.command !== 'versions'
+                      ? 'Project version. If running command in a CI environment and this option is not passed, the main project version will be used.'
+                      : 'A specific project version to view'
```
erunion, line 61:
```suggestion
                      : 'A specific project version to view.'
```

src/lib/versionOpt.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
erunion, line 1: Could add this to the base `Command` instead and then do `this.getVersionOption()` in `this.args` instantiations. We could do the same for the common `key` arg that most every command has too.

erunion (approved):

src/lib/versionOpt.js
kanadgupta: Yeah I thought about this but wasn't sure how to do so elegantly—some (but not all) commands use this opt, and then in some cases it uses a `version` opt with a different description:
https://github.com/readmeio/rdme/blob/2cd9847ad6c305f7e6db2e26148852d655ff9720/src/cmds/versions/index.ts#L43-L47

erunion: How about `this.getVersionOption('A specific project version to view')`> And then in the `getVersionOption` you can have a `description` type that's:

```ts
description: 'A specific project version to view' | 'the other description we use'
```

Could also make those descriptions an enum and then do `VersionOptionDescription.SPECIFIC` and `VersionOptionDescription.GENERAL`.

PR #563, opened 2022-08-10
feat: minor refactor of how we debug and log out data in commands

erunion:
## 🧰 Changes

Changes how we log out data in commands from calling `debug()` to having those commands set up as static methods in our core `Command` class.

> ℹ️  Why do `Command.debug()` instead of `this.debug()`? Unfortunately in a number of places we're logging out debug the `this` scope gets polluted to the inside of a nested function and `this.debug()` becomes undefined.


.eslintrc
```diff
@@ -30,15 +30,6 @@
      */
     "class-methods-use-this": ["error", { "exceptMethods": ["run"] }],
 
-    "import/order": ["error", {
```
erunion, line 33: Moved into our core standards and tagged a major release of that pkg. https://github.com/readmeio/standards/commit/edffdbe631ca3927b882fd19d8f03dd49451d937

kanadgupta (approved):

PR #564, opened 2022-08-10
feat: upgrading oas-normalize to pull in its new TS typings

erunion:
## 🧰 Changes

I rewrote [oas-normalize](https://npm.im/oas-normalize) last night in TypeScript (https://github.com/readmeio/oas-normalize/pull/210) so we no longer need to have a typing sinkhole exclusion for it.

### DIFF
--- a/src/lib/prepareOas.ts
+++ b/src/lib/prepareOas.ts
@@ -1,9 +1,8 @@
 import fs from 'fs';
 
 import chalk from 'chalk';
-import ora from 'ora';
-
 import OASNormalize from 'oas-normalize';
+import ora from 'ora';
 
 import { debug, info, oraOptions } from './logger';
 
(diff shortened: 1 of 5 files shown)


src/lib/prepareOas.ts
```diff
@@ -1,9 +1,8 @@
 import fs from 'fs';
 
 import chalk from 'chalk';
-import ora from 'ora';
-
 import OASNormalize from 'oas-normalize';
```
erunion, line 4: This wasn't being sorted before `ora` before because the sinkhole exclusion was causing the `import/order` rule to think that it was a local package.

src/lib/prepareOas.ts
```diff
@@ -69,7 +68,7 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
   let bundledSpec = '';
 
   if (command === 'openapi') {
-    bundledSpec = await oas.bundle().then((res: Record<string, unknown>) => {
```
erunion, line 72: Removed the type here because `OASNormalize.bundle` returns an `OpenAPI.Document` now and `Record` is incompatible with that.

kanadgupta: :hell-yeah:

kanadgupta (approved):

PR #565, opened 2022-08-10
ci: optimize test runs and add docs syncing dry run

kanadgupta:
## 🧰 Changes

- [x] Makes some tiny optimizations [our CI workflow `on` events](https://github.com/readmeio/rdme/blob/66791a6968d1e33fc21388be4b08313cd06dd85a/.github/workflows/ci.yml#L4-L7) so we're not running our CI twice for every push in a PR.
- [x] Performs a dry run of our docs syncing workflow on every push. This is so we can avoid what happened yesterday in [this workflow](https://github.com/readmeio/rdme/runs/7755364437?check_suite_focus=true) where our sync failed but we didn't realize it was merged into `main`.

## 🧬 QA & Testing

Are tests still running on every push to the PR, but with fewer duplicate workflow runs? And is [the docs syncing workflow](https://github.com/readmeio/rdme/runs/7775815962?check_suite_focus=true) running as expected and properly skipping over the actual syncing step?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   build:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,10 +1,7 @@
 name: Sync `documentation` directory to ReadMe
 
-# Run workflow for every push to the `main` branch
-on:
-  push:
-    branches:
-      - main
+# Run workflow for every push
+on: push
 
 jobs:
   sync:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
erunion, line 7: Does this get fired for every push to a pull request or just when one's created?

erunion (approved):

.github/workflows/ci.yml
kanadgupta: Both! Here's a snippet of the default `pull_request` behavior from [these docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):

> By default, a workflow only runs when a pull_request event's activity type is `opened`, `synchronize`, or `reopened`.

As the name implies, the `opened` activity type is for whenever a PR is opened.

Here's a description of `synchronize` from [these docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request):

> Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed.

PR #567, opened 2022-08-15
feat(openapi): auto-detect API definitions

kanadgupta:
| 🚥 Fix RM-5059 |
| :-- |

## 🧰 Changes

This swaps out our auto-detection of `swagger.json`, `openapi.yml`, etc. in the `openapi` and `validate` commands for a search functionality that's far more comprehensive.

`rdme` will now look for any and all JSON or YAML files (including in sub-directories) that contain a top-level [`openapi`](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields) or [`swagger`](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#fixed-fields) property.[^1] It will then prompt the user (if in a non-CI environment) if multiple files are found.

Here's what the new workflow looks like:

https://user-images.githubusercontent.com/8854718/184940170-b2005f46-6a58-4123-8213-a0593efc5cdb.mov

- [x] Implemented a new search functionality including support for `.gitignore` and a prompt-based system driven by [`prompts`](https://npm.im/prompts)
  - Will be migrating the rest of our `enquirer` usage over to `prompts` in a follow-up PR, ticketed in RM-5116
- [x] [Updated our docs accordingly](https://github.com/readmeio/rdme/pull/567/commits/e2332b3dfb358d3fdb779a3234d2fb9c19d95f47) (including some [minor unrelated touch-ups](https://github.com/readmeio/rdme/pull/567/commits/3adb2e321883aa60cc50ed18f05285b55675e185)) 📝 
- [x] Added a bunch of test coverage 🧪 

## 🧬 QA & Testing

Added a whole lotta test coverage which is now passing ✅ 

Also confirmed that this works when run locally (including in different directories via `npm link`). I added `2.0/` to the `.gitignore` file in the [`oas-examples`](https://github.com/readmeio/oas-examples) repo and confirmed that this is what I got when running a linked version of `rdme` in the repo:

https://user-images.githubusercontent.com/8854718/184941555-ff088fc0-3cb2-4bfa-b5da-4d177d3fccb1.mov

[^1]: `rdme` will not scan anything in the working directory's `.git/` directory, nor any files/directories specified in the `.gitignore` file (if the file exists).


src/lib/prepareOas.ts
```diff
@@ -18,28 +27,94 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
...
+    const action = command === 'openapi' ? 'upload' : 'validate';
+
+    const ignoreFilter = ignore().add('.git/');
+
+    if (fs.existsSync('.gitignore')) {
+      debug('.gitignore file found, adding to ignore filter');
+      ignoreFilter.add(fs.readFileSync('.gitignore').toString());
+    }
+
+    const jsonAndYamlFiles = readdirRecursive('.', ignoreFilter.createFilter()).filter(
```
erunion, line 56: You might want to integrate the `.gitignore` lookup detection into this method directly so that when it recurs into a new directory it looks for a `.gitignore` there. I don't think we can always assume that `.gitignore` will be in the root.

src/lib/prepareOas.ts
```diff
@@ -18,28 +27,94 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
...
+              debug(`error extracting OpenAPI/Swagger version for ${file}: ${e.message}`);
+              return '';
+            });
+        })
+      )
+    ).filter(Boolean);
+
+    debug(`number of possible OpenAPI/Swagger files found: ${possibleSpecFiles.length}`);
+
+    if (possibleSpecFiles.length === 0) {
```
erunion, line 86:
```suggestion
    if (!possibleSpecFiles.length) {
```

src/lib/prepareOas.ts
```diff
@@ -18,28 +27,94 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
...
+    }
+
+    specPath = possibleSpecFiles[0];
+
+    if (possibleSpecFiles.length === 1) {
+      fileFindingSpinner.succeed(`${fileFindingSpinner.text} found! 🔍`);
+      info(chalk.yellow(`We found ${specPath} and are attempting to ${action} it.`));
+    }
+
+    if (possibleSpecFiles.length > 1) {
```
erunion, line 100:
```suggestion
    } else if (possibleSpecFiles.length > 1) {
```

src/lib/prepareOas.ts
```diff
@@ -18,28 +27,94 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
...
+      /* istanbul ignore next */
+      if ((ciDetect() && process.env.NODE_ENV !== 'testing') || process.env.TEST_CI) {
```
erunion, line 102: Do we still need to check `NODE_ENV` if we have the `TEST_CI` env var?

src/lib/readdirRecursive.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param {String} folderToSearch path to directory
+ * @param pathFilter a function that takes in a directory/file path
+ * and returns a boolean. used for optimizing recursion paths
+ * e.g. paths specified in a file like `.gitignore`
+ * @link https://github.com/kaelzhang/node-ignore#createfilter
+ * @returns {String[]} array of files
+ */
+export default function readdirRecursive(
+  folderToSearch: string,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
erunion: You can fix this by changing the `pathFilter` arg to the following:

```
pathFilter: (path: string) => boolean = () => true
```

What you're defining now is the argument default, not its type. If you flesh out both this linting error will go away.

erunion (changes requested):

src/lib/prepareOas.ts
kanadgupta: Sadly yes, since `ciDetect()` returns true when running our tests in GitHub Actions and we want to only test this if-block in specific tests.

src/lib/readdirRecursive.ts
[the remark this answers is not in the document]
kanadgupta: ugh thank you so much 😭 I was trying to google what this should look like (i.e. defining the types and defaults for a function parameter) and was getting nowhere

src/lib/prepareOas.ts
kanadgupta: Added here (along with tests): https://github.com/readmeio/rdme/pull/567/commits/ac68c2ab9e4c1ce362ddc66e160b0efa2b4c40c7

src/lib/readdirRecursive.ts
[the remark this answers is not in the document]
kanadgupta: lmao refactored this function signature as part of https://github.com/readmeio/rdme/pull/567/commits/ac68c2ab9e4c1ce362ddc66e160b0efa2b4c40c7 so this was all for nothing 🤡

src/lib/prepareOas.ts
```diff
@@ -1,45 +1,107 @@
...
+        file.toLowerCase().endsWith('.json') ||
+        file.toLowerCase().endsWith('.yaml') ||
+        file.toLowerCase().endsWith('.yml')
+    );
+
+    debug(`number of JSON or YAML files found: ${jsonAndYamlFiles.length}`);
+
+    const possibleSpecFiles = (
+      await Promise.all(
+        jsonAndYamlFiles.map(async file => {
```
erunion, line 56: This need to be async?

```suggestion
        jsonAndYamlFiles.map(file => {
```

erunion (approved):
hell yeah

src/lib/prepareOas.ts
kanadgupta: ah damn left over from a previous refactor, nice catch

kanadgupta: wish TS or ESLint would catch this

erunion: yeah same, the rule we have in place only catches them on non-closure declarations. kind of annoying.

PR #570, opened 2022-08-17
refactor: migrate from `enquirer` to `prompts`

kanadgupta:
| 🚥 Fix RM-5116 |
| :-- |

## 🧰 Changes

Migrating our prompts system from the universally-disliked [`enquirer`](https://github.com/enquirer/enquirer) to [`prompts`](https://github.com/terkelg/prompts). In addition to it being a far better from a DX and testing standpoint, we also get a few nice helpful cues from `prompts` when a user is required to take an action (note the grayed out "use arrow-keys" guidance):

image

Overall, the prompt object patterns are fairly similar and there shouldn't be a whole lot of material changes for users (except in maybe the `login` command since that was previously using the ancient [`read`](https://npm.im/read) package).

### Which Commands Are Affected?

All areas of `rdme` that use a prompts-based system have been migrated over, namely:

- [x] Version selection (prompt: selecting project versions in the `categories`, `docs`, `openapi`, and `versions` command families)
- [x] `openapi` (prompt: selecting a spec)
- [x] `versions:create` (prompt: setting properties on the version you want to create)
- [x] `versions:update` (prompt: similar to above[^2])
- [x] `login` (prompt: logging in!)

### Other Small Cleanups 🧹 

Because this work got me looking into commands that don't get a lot of love, here are a bunch of other miscellaneous changes in this PR:
 - [x] I cleaned a bunch of types and added a bunch of tests in the `versions` group of commands (and honestly there's still a lot of work left to be done) and may or may not have fixed a few random bugs in the process. Not sure how much these commands are being used, I'd honestly love if we did away with them at some point 👀 
 - [x] Swapped out [`isemail`](https://npm.im/isemail) for [`vaildator`](https://npm.im/vaildator) since it's better maintained and a bit smaller of a dep.
 - [x] Shortened our debug command from `debug:bin` to just `debug`.
 - [x] Finally got our `openapi` command test coverage to 💯 
 - [x] Added a smol send off when someone exits the prompt system early:

image

## 🧬 QA & Testing

Tests (particularly for the version commands) should be far more resilient, so if those pass then that's a really good sign[^1]. I've tried a bunch of different user journeys locally (and backfilled test coverage where I easily could) and everything appears to work well!

[^1]: I still **cannot** get over how almost all of our `enquirer`-mocked tests were simply not testing anything at all for like... 3+ years now. Good riddance 🖕🏽 
[^2]: you'd think it was similar enough to `create` that this would be easy... but they're annoyingly different

PR #572, opened 2022-08-19
feat: creation of new `openapi:reduce` command to reduce large OpenAPI definitions

erunion:
| 🚥 Fix RM-5114 |
| :-- |

## 🧰 Changes

Earlier this week I ported the reduction functionality in [oas-reducer](https://npm.im/oas-reducer), a rad library I built last year to reduce an OpenAPI definition, over into [oas](https://npm.inm/oas) with the intention of porting the CLI side of it in here as a new command. This is that! Why? `oas-reducer` is cool as hell and deserves a larger audience[^1].

Thanks to the refactoring going on in #570 developing and testing this was a goddamned **dream**.

> ℹ️ After we eventually publish a new major release of `rdme` for all the work we've been doing I'll formally deprecate `oas-reducer` in NPM and archive the repo. 

## 🧬 QA & Testing

Here's what it looks like when you supply it a spec[^2]:

https://user-images.githubusercontent.com/33762/185554013-bb096135-bf81-4400-83da-977d2566f52a.mov

And with the work that's been happening with the `promptOas` library, and its file selection work, it looks great when you _don't_ give it a spec:

https://user-images.githubusercontent.com/33762/185554379-54af0314-ae3c-4dbc-aa3e-1b2295fd2289.mov

[^1]: It especially deserves to be core to `rdme` and the whole `rdme openapi` flow if CX or Sales occasionally recommend it to customers to chunk up huge OAS files into smaller ones so they can be uploaded within our smol Heroku timeouts.
[^2]: The error at the end of this video is because I didn't select a method to filter down that path by. I've since fixed this by adding a `min: 1` to the `methods` question.


__tests__/cmds/openapi/reduce.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  afterEach(() => {
+    consoleInfoSpy.mockRestore();
+
+    process.chdir(testWorkingDir);
+
+    jest.clearAllMocks();
+  });
+
+  describe('reducing', () => {
```
erunion, line 33: I'm not testing the meat of the reduction library here, just what the command is supplying it, because the reduction library has pretty decent test coverage over in the [`oas` repo](https://github.com/readmeio/oas/blob/main/__tests__/lib/reducer.test.ts).

__tests__/cmds/openapi/reduce.test.ts
```diff
@@ -0,0 +1,151 @@
...
+      it.each([
+        ['OpenAPI 3.0', 'json', '3.0'],
+        ['OpenAPI 3.0', 'yaml', '3.0'],
+        ['OpenAPI 3.1', 'json', '3.1'],
+        ['OpenAPI 3.1', 'yaml', '3.1'],
+      ])('should support reducing a %s definition (format: %s)', async (_, format, specVersion) => {
+        const spec = require.resolve(`@readme/oas-examples/${specVersion}/${format}/petstore.${format}`);
+
+        let reducedSpec;
+        fs.writeFileSync = jest.fn((f, d) => {
```
erunion, line 44: Cleaner to mock this out than writing, reading, and deleting the test file on the filesystem.

__tests__/__fixtures__/relative-ref-oas/petstore.json
```diff
@@ -106,6 +106,30 @@
           }
         ]
       }
+    },
+    "/user": {
```
erunion, line 110: Added this because I was using this spec to test tag reduction and I needed more than just the `pet` tag in here to make sure we didn't get back `user` from the reducer.

__tests__/lib/__snapshots__/commands.test.ts.snap
```diff
@@ -29,10 +29,15 @@ Object {
...
       },
+      Object {
+        "description": "Reduce an OpenAPI definition into a smaller subset.",
+        "name": "openapi:reduce",
+        "position": 3,
+      },
       Object {
         "description": "Alias for \`rdme openapi\`. [deprecated]",
         "name": "swagger",
-        "position": 2,
```
erunion, line 35: I'm surprised that this has always been showing up after `validate`  in `--help` with them having the same position.

src/cmds/openapi/index.ts
```diff
@@ -1,4 +1,4 @@
-import type { CommandOptions } from '../lib/baseCommand';
+import type { CommandOptions } from '../../lib/baseCommand';
```
erunion, line 1: Moved this into `src/cmds/openapi/index.ts` because the `openapi` command group is a family now. :corona: :vindiesel:

__tests__/cmds/openapi/reduce.test.ts
kanadgupta: writing out jest mocks

kanadgupta: jest mocks freak me out after #570 but this approach makes sense lol

__tests__/lib/__snapshots__/commands.test.ts.snap
kanadgupta: :lots-of-yikes:

src/cmds/openapi/index.ts
kanadgupta: oh the v8[^1] release notes are definitely going to be F&F-themed now

[^1]: also v8... multitudes!!!!

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+
+          return methods.map((method: string) => ({
+            title: method.toUpperCase(),
+            value: method,
+          }));
+        },
+      },
+      {
+        type: 'text',
+        name: 'ouputPath',
```
kanadgupta, line 119:
```suggestion
        name: 'outputPath',
```

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const reducedSpec = oasReducer(parsedBundledSpec, {
+      tags: promptResults.tags || [],
+      paths: (promptResults.paths || []).reduce((acc: Record<string, string[]>, p: string) => {
+        acc[p] = promptResults.methods;
+        return acc;
+      }, {}),
+    });
+
+    Command.debug(`saving reduced spec to ${promptResults.ouputPath}`);
```
kanadgupta, line 154:
```suggestion
    Command.debug(`saving reduced spec to ${promptResults.outputPath}`);
```

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    await fs.writeFileSync(promptResults.ouputPath, JSON.stringify(reducedSpec, null, 2));
```
kanadgupta, line 156: does this need to be `await`-ed?
```suggestion
    fs.writeFileSync(promptResults.outputPath, JSON.stringify(reducedSpec, null, 2));
```

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    Command.debug('reduced spec saved');
+
+    return Promise.resolve(chalk.green(`${specPath} has been reduced to ${promptResults.ouputPath}! 🤏`));
```
kanadgupta, line 160:
```suggestion
    return Promise.resolve(chalk.green(`${specPath} has been reduced to ${promptResults.outputPath}! 🤏`));
```

src/lib/prepareOas.ts
```diff
@@ -14,12 +14,12 @@ type FileSelection = {
 /**
  * Normalizes, validates, and (optionally) bundles an OpenAPI definition.
  *
- * @param {String} path path to spec file. if this is missing, the current directory is searched
- * for certain file names
- * @param command string to distinguish if it's being run in
- * an 'openapi' or 'validate' context
+ * @param path Path to a spec file. If this is missing, the current directory is searched for
+ *    certain file names.
+ * @param action The action context in which this is being run within (uploading a spec,
```
kanadgupta, line 19:
```suggestion
 * @param command The command context in which this is being run within (uploading a spec,
```

src/lib/prepareOas.ts
```diff
@@ -40,7 +40,17 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
 
     const fileFindingSpinner = ora({ text: 'Attempting to locate API definitions...', ...oraOptions() }).start();
 
-    const action = command === 'openapi' ? 'upload' : 'validate';
+    let action;
```
kanadgupta, line 43: TS noob question but is there any benefit to a stricter type here?

```suggestion
    let action: string;
```
or maybe even this:
```suggestion
    let action: 'upload' | 'reduce' | 'validate';
```

src/lib/prepareOas.ts
```diff
@@ -40,7 +40,17 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
 
     const fileFindingSpinner = ora({ text: 'Attempting to locate API definitions...', ...oraOptions() }).start();
```
kanadgupta, line 41: this one's on me lol but mind changing this too? in hindsight I hate this original language I wrote lol

```suggestion
    const fileFindingSpinner = ora({ text: 'Looking for API definitions...', ...oraOptions() }).start();
```

src/lib/prepareOas.ts
```diff
@@ -14,12 +14,12 @@ type FileSelection = {
...
- * for certain file names
- * @param command string to distinguish if it's being run in
- * an 'openapi' or 'validate' context
+ * @param path Path to a spec file. If this is missing, the current directory is searched for
+ *    certain file names.
+ * @param action The action context in which this is being run within (uploading a spec,
+ *    validation, or reducing one).
  */
-export default async function prepareOas(path: string, command: 'openapi' | 'validate') {
+export default async function prepareOas(path: string, command: 'openapi' | 'openapi:reduce' | 'validate') {
```
kanadgupta, line 22: ugh i'm SO excited that we're able to find yet another useful command that can piggyback off of the spec-finding logic 🥹

src/lib/prepareOas.ts
```diff
@@ -40,7 +40,17 @@ export default async function prepareOas(path: string, command: 'openapi' | 'val
...
+    switch (command) {
+      case 'openapi':
+        action = 'upload';
+        break;
+      case 'openapi:reduce':
+        action = 'reduce';
+        break;
+      default:
+        action = command;
+    }
```
kanadgupta, line 53: once we migrate `validate` over to `openapi:validate` we should be able to just dynamically strip away the `openapi:` prefixes :chefs-kiss-cowboy:

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+            title: method.toUpperCase(),
+            value: method,
+          }));
+        },
+      },
+      {
+        type: 'text',
+        name: 'ouputPath',
+        message: 'Enter the path to save your reduced API definition to:',
+        validate: value => {
```
kanadgupta, line 121: Thoughts on having a nice default here to save the user a click? Maybe something like `filename-reduced-timestamp.{json|yaml}`

```suggestion
        initial: 'some-nice-default.json',
        validate: value => {
```

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (specType !== 'OpenAPI') {
+      throw new Error('Sorry, this reducer feature in rdme only supports OpenAPI 3.0+ definitions.');
+    }
+
+    const promptResults = await promptTerminal([
+      {
+        type: 'select',
+        name: 'reduceBy',
+        message: 'Would you like to reduce by paths or tags?',
+        initial: 0,
```
kanadgupta, line 64: nit but I don't _think_ you need a default here, I believe the `select` prompt just selects the first one by default:
```suggestion
```

src/cmds/openapi/reduce.ts
kanadgupta, line 156: How long could the `oasReducer` + `fileSync` op take? Does it make sense to add another `ora` spinner for this part?

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,162 @@
...
+        choices: [
+          { title: 'Tags', value: 'tags' },
+          { title: 'Paths', value: 'paths' },
+        ],
+      },
+      {
+        type: (prev, values) => (values.reduceBy === 'tags' ? 'multiselect' : null),
```
kanadgupta, line 71: these `multiselect` prompts are soooo nice 😭

kanadgupta (changes requested):
overall looking good, most of my comments are minor! chatted briefly about this offline but I saw this in the tag selection (note the `[object Object]`):

image

src/cmds/openapi/reduce.ts
erunion: Wow and thanks to TS I made that typo literally everywhere.

src/cmds/openapi/reduce.ts
erunion: On a HUGE spec maybe a second or two, it's generally seemed to be pretty quick.

src/lib/prepareOas.ts
erunion: It's set up as an `any` but because we're filling it with strings when it's used its implied type ends up being a string:

Screen Shot 2022-08-19 at 9 32 30 AM

I like the more specific type union option though.

src/cmds/openapi/reduce.ts
kanadgupta: mm don't think it hurts, what do you think? i err on the side of more vs. less loading UI so the user doesn't think the CLI is bricked

src/cmds/openapi/reduce.ts
erunion: Added in a dynamic initial:

Screen Shot 2022-08-19 at 9 57 15 AM

erunion:
> chatted briefly about this offline but I saw this in the tag selection (note the `[object Object]`):
> 
> image

Turns out this was because that spec has a `{name: 'tag'}` XML property in it and that was being picked up by the loose JSONPath query. I've added some filtering to filter these out:

Screen Shot 2022-08-19 at 10 39 06 AM

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,189 @@
...
+          }));
+        },
+      },
+      {
+        type: 'text',
+        name: 'outputPath',
+        message: 'Enter the path to save your reduced API definition to:',
+        initial: () => {
+          const extension = path.extname(specPath);
+          return `${path.basename(specPath).split(extension)[0]}-reduced${extension}`;
```
erunion, line 132: This'll dynamically turn `petstore.json` or `petstore.yaml` into `petstore-reduced.{ext}`

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,189 @@
...
+        message: 'Choose which tags to reduce by:',
+        min: 1,
+        choices: () => {
+          const tags = jsonpath
+            .query(parsedBundledSpec, '$..paths..tags')
+            .flat()
+            .filter(tag => {
+              // A spec may have a `name: tag` which this JSONPath query will pick up. Since that
+              // definitely isn't a tag we want here we need to filter them out.
+              return typeof tag === 'string';
```
kanadgupta, line 84: damn nice find... I wish there was an easy way to test this but can't really think of one 🤔

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,189 @@
...
+        paths: (promptResults.paths || []).reduce((acc: Record<string, string[]>, p: string) => {
+          acc[p] = promptResults.methods;
+          return acc;
+        }, {}),
+      });
+
+      spinner.succeed(`${spinner.text} done! ✅`);
+    } catch (err) {
+      Command.debug(`err=${err.message}`);
+      throw err;
```
erunion, line 176: The reducer throws exceptions for errors (eg. if you try to filter out too much) right now so instead of trying to capture and put those in some sort of "Yikes! We couldn't reduce this because...` error I'm just letting it do its thing.

src/cmds/openapi/reduce.ts
```diff
@@ -0,0 +1,189 @@
...
+        tags: promptResults.tags || [],
+        paths: (promptResults.paths || []).reduce((acc: Record<string, string[]>, p: string) => {
+          acc[p] = promptResults.methods;
+          return acc;
+        }, {}),
+      });
+
+      spinner.succeed(`${spinner.text} done! ✅`);
+    } catch (err) {
+      Command.debug(`err=${err.message}`);
```
kanadgupta, line 175: might wanna kill the spinner in this block too:
```suggestion
      Command.debug(`reducer err: ${err.message}`);
      spinner.fail();
```

src/cmds/openapi/reduce.ts
kanadgupta: :nice-rgb:

kanadgupta (approved):
tiny nit but otherwise LGTM!

src/cmds/openapi/reduce.ts
erunion: Yeah I'm not sure how to test what a prompt in `prompts` is rendering without factoring this work out into a separate small utility.

Issue #568, opened 2022-08-16
Use --version even if --id is provided


sofianegargouri:
If I'm not providing `--id`, I can update a specific version of my project, but with an input in the CLI:
```
➜ rdme openapi doc/openapi.yaml \
               --key=MYKEY \
               --version=0.1
✔ Validating API definition located at doc/openapi.yaml... done! ✅
✔ Staging your API definition for upload... done! 🚀
✔ Would you like to update an existing OAS file or create a new one? · update
✔ Select your desired file to update · MYID
✔ Updating your API docs in ReadMe... done! 🦉
You've successfully updated an existing OpenAPI file on your ReadMe project!

        https://dash.readme.com/project/project_slug/v0.1/refs/myref
```

But if I'm providing the `--id`, the `--version` argument is ignored so I can't change the target version I want to update:
```
➜ rdme openapi doc/openapi.yaml \
               --key=MY_KEY \
               --version=0.1 \
               --id=MY_ID
⚠️  Warning! We'll be using the version associated with the `--id` option, so the `--version` option will be ignored.
✔ Validating API definition located at doc/openapi.yaml... done! ✅
✔ Staging your API definition for upload... done! 🚀
✔ Updating your API docs in ReadMe... done! 🦉
You've successfully updated an existing OpenAPI file on your ReadMe project!

        https://dash.readme.com/project/project_slug/v1.0/refs/myref
```
As you can see in the recap link, it did update the `1.0` version and not the `0.1`.

I have a workaround to work locally, but this behavior is annoying for CIs

kanadgupta:
Hi @sofianegargouri — note the warning output in that second command:

```
⚠️  Warning! We'll be using the version associated with the `--id` option, so the `--version` option will be ignored.
```

This is by design of the ReadMe data model—API definition IDs roll up to project versions (i.e. API definition IDs have a many-to-one relationship with project versions). The version is therefore inferred from the API definition ID, which is why the `--version` parameter is ignored.

In the output of that first command you wrote (i.e. the one with the `--version` parameter but not with the `--id` parameter), there should be a sample command in the output that you can run so it updates the API definition and version you're expecting. It should look like this:

```
To update your OpenAPI definition, run the following:

    rdme openapi doc/openapi.yaml --key=<key> --id=YOUR_ID
```

Hope that helps!

sofianegargouri:
I didn't get the fact that the ID changes depending on the version I'm updating. Got that now, thanks a lot !

PR #575, opened 2022-08-19
feat(openapi): add `--create` flag

kanadgupta:
| 🚥 Fix RM-5148 |
| :-- |

## 🧰 Changes

Adds a `--create` flag to the `openapi` command so users can bypass the prompt that asks them whether or not they want to create or update a spec.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/src/cmds/openapi/index.ts
+++ b/src/cmds/openapi/index.ts
@@ -92,6 +92,10 @@ export default class OpenAPICommand extends Command {
       );
     }
 
+    if (create && id) {
+      Command.warn("We'll be using the `--create` option , so the `--id` parameter will be ignored.");
+    }
+
     // Reason we're hardcoding in command here is because `swagger` command
     // relies on this and we don't want to use `swagger` in this function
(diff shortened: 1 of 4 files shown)


src/cmds/openapi/index.ts
```diff
@@ -86,6 +92,10 @@ export default class OpenAPICommand extends Command {
       );
     }
 
+    if (create && id) {
+      Command.warn("We'll be using the `--create` option , so the `--id` parameter will be ignored.");
```
erunion, line 96: Wonder if in this case we should have an "are you sure you want to do this" prompt or maybe them manually supplying the create arg is enough consent.

erunion (approved):

src/cmds/openapi/index.ts
kanadgupta: yeah I went back and forth on this and figured that the `--id` parameter causes some confusion (see https://github.com/readmeio/rdme/issues/568) whereas the word `create` should hopefully be intuitive enough. If we get negative feedback then we can revisit.

PR #576, opened 2022-08-19
chore(prompt): various enhancements and refactors

kanadgupta:
| 🚥 Fix RM-3621 (and a bunch of other stuff) |
| :-- |

## 🧰 Changes

- [x] Refactored the `login` command to use [overrides](https://github.com/terkelg/prompts#override) and added a few tests
- [x] Refactored our [`onState`](https://github.com/terkelg/prompts#onstate) usage to use [`onCancel`](https://github.com/terkelg/prompts#optionsoncancel) (see 16f61172839e1ced3b0c9bd8a555416dca967bad for more reasoning)
- [x] Removed the apparently unnecessary `enableTerminalCursor` function. Not sure if it was working before but I've tried a bunch of things and can't get my cursor to disappear. I'm cool with keeping it in there just to be safe too, let me know what you think 🤷🏽 
- [x] Refactored our authenticated commands so we dynamically detect the `key` argument via the `args` array, rather than having a flag that we have to manually enable
- [x] Added top-level CI detection so our prompts error out in CI environments

## 🧬 QA & Testing

Tests appear to pass, but it might be good to check out this branch, run the build, and try messing around.

- [ ] Does the terminal cursor remain intact no matter what commands you try?
- [ ] Are you still able to safely exit out of prompts using <kbd>CTRL</kbd>+<kbd>C</kbd> or <kbd>ESC</kbd>?
- [ ] Does CI detection work for you properly?[^1]

[^1]: You can test this by prefixing any command with a [CI env variable](https://github.com/npm/ci-detect/blob/main/lib/index.js). I've been running `GITLAB_CI=true bin/rdme login`

### DIFF
--- a/src/cmds/login.ts
+++ b/src/cmds/login.ts
@@ -2,6 +2,7 @@ import type { CommandOptions } from '../lib/baseCommand';
 
 import chalk from 'chalk';
 import config from 'config';
+import prompts from 'prompts';
 import isEmail from 'validator/lib/isEmail';
 
 import Command, { CommandCategories } from '../lib/baseCommand';
@@ -44,9 +45,9 @@ export default class LoginCommand extends Command {
   async run(opts: CommandOptions<Options>) {
     super.run(opts);
 
-    let { project } = opts;
+    prompts.override(opts);
 
-    const promptResults = await promptTerminal([
+    const { email, password, project, token } = await promptTerminal([
       {
         type: 'text',
         name: 'email',
@@ -62,7 +63,7 @@ export default class LoginCommand extends Command {
         message: 'What is your password?',
       },
       {
-        type: opts.project ? null : 'text',
+        type: 'text',
         name: 'project',
         message: 'What project are you logging into?',
         initial: configStore.get('project'),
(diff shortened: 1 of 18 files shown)


src/cmds/login.ts
```diff
@@ -44,9 +45,9 @@ export default class LoginCommand extends Command {
   async run(opts: CommandOptions<Options>) {
     super.run(opts);
 
-    let { project } = opts;
+    prompts.override(opts);
```
kanadgupta, line 48: I got irrationally excited about this one lol

erunion (approved):
can't replicate the missing cursor problem with this but that auth refactor is 💯

src/cmds/login.ts
kanadgupta: yeahhh I read about override and login was the first thing I thought of... I don't want to look at the `versions` commands for a long time lmao but I looked into doing this for the `prepareOas` handler and we have enough complexity to our spec searching that I can't think of a useful refactor approach

erunion: yeah i'd rather remove the versions commands before even thinking about doing anything else to them

PR #577, opened 2022-08-22
feat(GHA): workflow file creation onboarding

kanadgupta:
| 🚥 Fix RM-5048 |
| :------------- |

## 🧰 Changes

The Linear ticket has all the details on what we're trying to do here!

### Must Haves:

- [x] GitHub Action Onboarding + Codegen ✏️ 
  - [x] `validate`
  - [x] `openapi`
  - [x] `docs` (and friends[^2])
- [x] YAML file tests
  - [x] `validate`
  - [x] `openapi`
  - [x] `docs` (and friends[^2])
- [x] E2E tests
  - [x] `validate`
  - [x] `openapi`
  - [x] `docs`
- [x] Magical repo detection ✨ 
- [x] Better error handling for if user doesn't have `.github/workflows` directory created
- [x] Save a little boolean to the `configstore` when we ask if they'd like to set it up so we annoy them with too many asks
  - [x] Major release matching (h/t @erunion for the idea)
- [x] Link out to [GitHub's docs on adding repo secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) (that includes a little snippet of the API key)
  - [x] And just an overall pass at all the language
- [ ] Docs (to happen in a separate PR, ticketed in RM-5060) 

### Nice to haves:

- [x] Validation of the GitHub workflows using [the JSON Schema definition](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-workflow.json)

#### Archive
- Automatic [dependabot file creation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)
  - [Dependabot schema validation](https://json.schemastore.org/dependabot-2.0.json) maybe?
- Link out to the GitHub repo settings page for repo secrets creation[^1]

## 🧬 QA & Testing

I've added a metric ton of tests for all of this, but I think once this is merged, we should play around with it in a beta release just to make sure it functions as expected.

If you'd like to play around with this work yourself, check out this branch and run the following commands:

```sh
# Install and set up the build
npm ci && npm run build

# You can use this command and select a valid file from the prompt, like ` __tests__/__fixtures__/petstore-simple-weird-version.json`
bin/rdme validate

# Alternatively, you can just pass the `github` flag which will run the GHA creation flow every time:
bin/rdme validate --github

# Alternatively, you can just pass the file directly into the command:
bin/rdme validate __tests__/__fixtures__/petstore-simple-weird-version.json
```

I used `validate` above since that one is the quickest to test (i.e. doesn't require auth), but you can also test any of the following commands as well:
- `openapi`
- `docs`
- `docs:single`
- `changelogs`
- `changelogs:single`
- `custompages`
- `custompages:single`

If you follow the prompts, you'll end up with a brand-spankin'-new GitHub Actions workflow file that contains that command 🪄 

[^1]: (this seems like it could be very prone to breakage... not gonna do it but I'll list it here for posterity)
[^2]: `changelogs`, `custompages`, etc... [it's fairly easy to add new commands](https://github.com/readmeio/rdme/pull/577#discussion_r954388701)! I'll probably just write E2E tests for the `docs` command since the other ones are just super lightweight wrappers around the `docs` logic


src/lib/createGHA/index.ts
```diff
@@ -0,0 +1,234 @@
...
+    containsNonGitHubRemote = /gitlab/.test(remotesList) || /bitbucket/.test(remotesList);
+  }
+
+  const repoRoot = await git.revparse(['--show-toplevel']).catch(e => {
+    debug(`error grabbing git root: ${e.message}`);
+    return '';
+  });
+
+  return { containsGitHubRemote, containsNonGitHubRemote, defaultBranch, isRepo, repoRoot };
+}
```
kanadgupta, line 103: @RyanGWU82 here's where we're running those git commands! The idea is that we want to automatically determine if we should ask the user if they'd like to generate a new GitHub Action workflow. See here for the business logic!

In order to do this, we'll need a few bits of information:

| Data We Need | Corresponding `git` command |
| :-------------: | :-------------:
| Flag to indicate if the repo is connected to GitHub | `git remote -v` |
| Flag to indicate if the repo is connected to **a non-GitHub remote** | `git remote -v` |
| What the default branch for the repo is | `git remote` and then `git remote show origin`[^1] |
| What the root for the repo is in the current directory | `git rev-parse --show-toplevel` |

Any and all feedback on this approach would be appreciated 😇 

[^1]: `origin` in the second command is pulled dynamically from the first line of the output of the first command, and then the `HEAD` branch is grabbed via some regex 😬

src/lib/createGHA/index.ts
```diff
@@ -0,0 +1,234 @@
...
+      isCI() ||
+      // is a repo, but only contains non-GitHub remotes
+      (isRepo && containsNonGitHubRemote && !containsGitHubRemote) ||
+      // not testing this function
+      (process.env.NODE_ENV === 'testing' && !process.env.TEST_CREATEGHA)
+    ) {
+      // We return the original command message and pretend this command flow never happened.
+      return msg;
+    }
+  }
```
kanadgupta, line 131: The tl;dr here is that we should only ask the user (if they're running `rdme` in a non-CI environment) if they want to create a GitHub Action workflow file if they meet any of the following criteria:
- They passed the `--github` flag (duh)
- They have a git repo set up that either has no remotes or a GitHub remote

src/lib/createGHA/index.ts
```diff
@@ -0,0 +1,234 @@
...
+   * By using `info` as opposed to `log`, we also can mock it in our tests
+   * while also freely using `log` when debugging our code.
+   *
+   * @see {@link https://github.com/readmeio/rdme/blob/main/CONTRIBUTING.md#usage-of-console}
+   */
+  // eslint-disable-next-line no-console
+  if (msg) console.info(msg);
+
+  // eslint-disable-next-line no-console
+  console.info('\n🐙 GitHub Repository detected! 🐙\n');
```
kanadgupta, line 146: Will be adding a better message here that explains the benefits, etc.

src/lib/createGHA/index.ts
```diff
@@ -0,0 +1,234 @@
...
+  const { output } = await transcludeFile('./src/lib/createGHA/base.yml', { resolvers: [customResolver] });
+
+  fs.writeFileSync(filePath, output);
+
+  return Promise.resolve(
+    [
+      chalk.green('\nYour GitHub Action has been created! ✨\n'),
+      "You're almost finished! Just a couple more steps:",
+      `1. Commit and push your newly created file (${filePath}) to GitHub 🚀`,
+      `2. Create a GitHub secret called \`${GITHUB_SECRET_NAME}\` and populate that with a ReadMe API key 🔑`,
```
kanadgupta, line 231: This success message will also be improved upon!

src/lib/createGHA/base.yml
```diff
@@ -0,0 +1,23 @@
+# This GitHub Action workflow was auto-generated by the `rdme` cli on :[](timestamp)
```
kanadgupta, line 1: We have this base file and we're using [`hercule`](npmjs.com/package/hercule) to populate it dynamically ✨

__tests__/lib/createGHA.test.ts
```diff
@@ -0,0 +1,218 @@
...
+        });
+
+        await expect(createGHA('', cmd, command.args, opts)).resolves.toMatchSnapshot();
+
+        expect(yamlOutput).toBeValidSchema(ghaWorkflowSchema);
+        expect(yamlOutput).toMatchSnapshot();
+        expect(fs.writeFileSync).toHaveBeenCalledWith(`.github/workflows/${fileName}.yml`, expect.any(String));
+        expect(console.info).toHaveBeenCalledTimes(1);
+        const output = getCommandOutput();
+        expect(output).toMatch('GitHub Repository detected!');
```
kanadgupta, line 95: For every command that this can be run with, we're running a **ton** of tests:
- YAML file snapshot tests
- Command output snapshot tests
- `fs.writeFileSync` mock tests to ensure it's being called properly
- JSON Schema validation of the YAML output to ensure it conforms to the GitHub Action Workflow specification (see `__tests__/helpers/jest.matchers.ts`)

kanadgupta (commented):
@erunion @RyanGWU82 tagging y'all for an initial review before I'm OOO for a few days! Check out the (half-baked) PR description for an overview of the outstanding work.

There's a lot going on here so I left some comments with the highlights. Would love to hear your thoughts on the approach thus far, as well as the outstanding work and which tasks you think I should do vs. which ones are overkill! Thanks in advance 💙

src/lib/createGHA/index.ts
RyanGWU82: This all sounds great — I didn't even think it was possible to find the default branch on GitHub from a checkout.

"Whether or not this is a git-enabled repo" can be answered with `git rev-parse --show-toplevel` — if that succeeds then you're in a git-enabled repo. (Although relying on simple-git sounds fine too!)

(Haven't looked through the code yet) Are you using simple-git for all of this? Looks like it shells out to `git` on the user's system and it doesn't require a minimum git version? It might be a good idea to check to see if they have at least a somewhat modern version, just to avoid things going too badly…

RyanGWU82: You might also want to just confirm that the `git` command is available locally and give a friendly error if it's not, or catch any/all `simple-git` errors and replace them with friendlier messages.

src/lib/createGHA/index.ts
```diff
@@ -0,0 +1,258 @@
...
+        initial: true,
+      },
+      {
+        message: 'What branch on GitHub should this workflow run on?',
+        name: 'branch',
+        type: 'text',
+        initial: defaultBranch || 'main',
+      },
+      {
+        message: 'What would you like to name this file?',
```
RyanGWU82, line 185: Might be helpful to say something like "What would you like to name this GitHub Action?" and make it clear that the  file will be added in `.github/workflows`.

RyanGWU82:
Thanks for sending this along in a draft state. I really like how you've structured this so that it can handle all kinds of `rdme` commands. Very cool!

src/lib/createGHA/index.ts
kanadgupta: > "Whether or not this is a git-enabled repo" can be answered with git rev-parse --show-toplevel — if that succeeds then you're in a git-enabled repo. (Although relying on simple-git sounds fine too!)

I believe [`isRepo()`](https://github.com/steveukx/git-js#git-rev-parse--repo-properties) is using this! But yeah I figured it was just easier to use `simple-git` since it's basically just Node syntactical sugar for `git` commands.

> You might also want to just confirm that the git command is available locally and give a friendly error if it's not, or catch any/all simple-git errors and replace them with friendlier messages.

For some add'l context, all of this data is purely optional and for the purpose of enhancing the onboarding experience (e.g. to determine if the working directory a GitHub repo or not), so I wrote [the code](https://github.com/readmeio/rdme/blob/06895f474a79edd68b38efbd28dd4a6b7f12ad74/src/lib/createGHA/index.ts#L80-L119) in such a way that everything falls back to falsy values that we just ignore. Say a user is not in a repo or has an outdated version of `git`, we just hit [this conditional](https://github.com/readmeio/rdme/blob/06895f474a79edd68b38efbd28dd4a6b7f12ad74/src/lib/createGHA/index.ts#L137) and [exit the function](https://github.com/readmeio/rdme/blob/06895f474a79edd68b38efbd28dd4a6b7f12ad74/src/lib/createGHA/index.ts#L147-L148) 🤷🏽‍♂️

src/lib/createGHA/index.ts
kanadgupta: Good call! The confirmation message includes the full path including the `.github/workflows` directory, but I'll make this question a little bit more verbose so the user knows what exactly this file is.

src/lib/createGHA/baseFile.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * The base GitHub Actions workflow file, which is used for generating the YAML file.
+ *
+ * The reason this is expressed as a JavaScript string as opposed to a YAML file is because
```
erunion, line 4: Not totally accurate. You could add `path/to/file.yaml` into the `include` array in `tsconfig.json` and it'll get moved into `dist/` when you run `npm run build`. At least I think so!

erunion (commented):
giphy-1

src/cmds/changelogs/index.ts
```diff
@@ -68,6 +66,8 @@ export default class ChangelogsCommand extends Command {
       })
     );
 
-    return chalk.green(updatedDocs.join('\n'));
+    return Promise.resolve(chalk.green(updatedDocs.join('\n'))).then(msg =>
+      createGHA(msg, this.command, this.args, opts)
```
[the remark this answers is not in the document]
kanadgupta: Love this idea 😻

__tests__/cmds/openapi/reduce.test.ts
```diff
@@ -17,8 +17,6 @@ const getCommandOutput = () => consoleInfoSpy.mock.calls.join('\n\n');
 
 describe('rdme openapi:reduce', () => {
   beforeEach(() => {
-    jest.mock('fs');
```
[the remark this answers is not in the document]
kanadgupta: Yeah I was looking for `fs` mock examples for my tests and I noticed this. I don't think it's necessary since we're also doing this: https://github.com/readmeio/rdme/blob/3d081794d6bb17dff11aef274a46f6e727ec33ca/__tests__/cmds/openapi/reduce.test.ts#L42-L45

Tests appear to pass and the mocks appear to operate as expected 🤷🏽‍♂️

src/lib/createGHA/baseFile.ts
kanadgupta: Yeah I had tried this and the `include` array only supports compiled (i.e. JS/TS/etc.) files 😕 see this issue I linked to: https://github.com/microsoft/TypeScript/issues/30835

__tests__/lib/createGHA.test.ts
```diff
@@ -0,0 +1,218 @@
...
+describe('#createGHA', () => {
+  let ghaWorkflowSchema;
+
+  beforeAll(async () => {
+    ghaWorkflowSchema = await fetch(ghaWorkflowUrl)
+      .then(res => res.json())
+      .catch(() => {
+        console.error('error fetching JSON schema');
+        return ghaWorkflowSchemaBackup;
+      });
```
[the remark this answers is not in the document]
kanadgupta: Removed in https://github.com/readmeio/rdme/pull/577/commits/f88459f6d84b6ea7b50a420d12a52104afcc575f

src/lib/createGHA/base.yml
kanadgupta: (this is now outdated, see https://github.com/readmeio/rdme/commit/f03902b15b4bac663ddc8460af808f1f1e51ece9)

__tests__/lib/createGHA.test.ts
```diff
@@ -0,0 +1,306 @@
...
+        'custompages' as keyof typeof commands,
+        CustomPagesCommand,
+        { key, folder: './custompages' } as CommandOptions<{}>,
+      ],
+      [
+        'custompages:single' as keyof typeof commands,
+        SingleCustomPageCommand,
+        { key, filePath: './custompages/rdme.md' } as CommandOptions<{}>,
+      ],
+    ])('%s', (cmd, CmdClass, opts) => {
```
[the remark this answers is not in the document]
kanadgupta: Refactored in 050424d66ea2007d6c43429a7bd4f133ff1b7efe

__tests__/lib/createGHA.test.ts
```diff
@@ -0,0 +1,306 @@
...
+    git.checkIsRepo = jest.fn(() => {
+      return Promise.resolve(true) as unknown as Response<boolean>;
+    });
+
+    git.remote = getGitRemoteMock();
+
+    fs.writeFileSync = jest.fn((f, d) => {
+      yamlOutput = d;
+      return true;
+    });
```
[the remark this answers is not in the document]
kanadgupta: Consolidated in 0f956f902dbae244cd5225c091d734d6ace9128b

src/cmds/custompages/index.ts
```diff
@@ -12,6 +13,7 @@ export type Options = {
   folder?: string;
 };
 
+@supportsGHA
```
kanadgupta, line 16: The diffs in these files are way cleaner now @erunion :chefs-kiss:

erunion (approved):
I'm happy with this! Would love to encapsulate the `getGithubArg` stuff into the decorator at some point (because we can probably do a `@authRequired` one too) but I don't think it's worth the `--help` refactor it's going to require right now.

__tests__/lib/createGHA.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      { cmd: 'docs:single', CmdClass: SingleDocCommand, opts: { key, filePath: './docs/rdme.md', version: '1.0.0' } },
+      { cmd: 'changelogs', CmdClass: ChangelogsCommand, opts: { key, folder: './changelogs' } },
+      { cmd: 'changelogs:single', CmdClass: SingleChangelogCommand, opts: { key, filePath: './changelogs/rdme.md' } },
+      { cmd: 'custompages', CmdClass: CustomPagesCommand, opts: { key, folder: './custompages' } },
+      {
+        cmd: 'custompages:single',
+        CmdClass: SingleCustomPageCommand,
+        opts: { key, filePath: './custompages/rdme.md' },
+      },
+    ])('$cmd', ({ cmd, CmdClass, opts }) => {
```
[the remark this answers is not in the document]
kanadgupta: yeah i also learned about this today 😌

__tests__/cmds/docs/index.test.ts
```diff
@@ -411,4 +416,197 @@ describe('rdme docs', () => {
...
+    let yamlOutput;
+
+    const getCommandOutput = () => {
+      return [consoleInfoSpy.mock.calls.join('\n\n')].filter(Boolean).join('\n\n');
+    };
+
+    beforeEach(() => {
+      consoleInfoSpy = jest.spyOn(console, 'info').mockImplementation();
+
+      before((f, d) => {
```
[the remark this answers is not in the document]
kanadgupta: done in https://github.com/readmeio/rdme/pull/577/commits/a72f9ae8a56ea3edf424de8dcb9c560516f3601a! also made a similar change in `__tests__/cmds/openapi/reduce.test.ts`.

src/cmds/changelogs/index.ts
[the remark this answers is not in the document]
kanadgupta: done! https://github.com/readmeio/rdme/pull/577/commits/41883329928ab925b0aea87a64338d4b159f2eed

PR #583, opened 2022-08-29
feat(docs): Add id to success message response for `docs` and `docs:single`

garrett-wade:
| 🚥 Fix RM-XXX |
| :-- |

## 🧰 Changes

Added the id of a document to the success response when a document is uploaded. We often have developer publishing a single doc that will be used as the parent for another doc, and getting the id directly in the response improves the experience for the developer.

## 🧬 QA & Testing

Run the `docs:single` or `docs` command on a file you wish to upload.
Notice the id is added to the success message(s) printed in the console.

### DIFF
--- a/__tests__/cmds/changelogs/index.test.ts
+++ b/__tests__/cmds/changelogs/index.test.ts
@@ -195,6 +195,7 @@ describe('rdme changelogs', () => {
   describe('new changelogs', () => {
     it('should create new changelog', async () => {
       const slug = 'new-doc';
+      const id = '1234';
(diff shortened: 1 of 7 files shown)


kanadgupta (changes requested):
Thanks for the PR @garrett-wade—this change makes a lot of sense! One small copy suggestion below, otherwise LGTM!

erunion (dismissed):
can you take a look at the tests again here @garrett-wade? looks like they're all failing with that message change.

kanadgupta:
Thanks @garrett-wade!

kanadgupta (approved):

PR #584, opened 2022-08-30
test(`openapi`): fix a few incorrect mocks

kanadgupta:
## 🧰 Changes

As part of #577, I noticed an issue with a bunch of our `openapi` command tests where the response from [our `GET /version/{versionId}` endpoint](https://docs.readme.com/reference/getversion) returns a single version object but our mocks incorrectly have them returning an array of objects.

Even worse, the tests were passing just fine because we weren't checking for the version header in subsequent mocks 😬 it should contain a version but that header has always been undefined due to this discrepancy.

This PR cleans up the mess a bit and adds a few stricter mock checks.

## 🧬 QA & Testing

Did a quick search and confirmed that this issue is thankfully only affecting this test bed. Do tests pass?


erunion (approved):
woof

PR #597, opened 2022-09-06
feat: `npm init` alias

kanadgupta:
## 🧰 Changes

Per [the `npm init` docs](https://docs.npmjs.com/cli/v8/commands/npm-init), we can utilize the `@readme` namespace to do fancy things like:

```sh
npm init @readme
```

which would run the following:

```sh
npx @readme/create
```

This PR creates that `@readme/create` repo and aliases it to `rdme openapi --github`, but we might wanna do something fancier at some point. Also should this be versioned the same as `rdme`? Lots of open questions, but figured I'd open this PR to get the conversation started.

### DIFF
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -2,7 +2,7 @@
   "name": "@readme/create",
   "version": "0.0.1",
   "description": "Sync your OpenAPI definition to ReadMe",
-  "bin": "npx rdme@latest openapi --github",
+  "bin": "npx rdme@next openapi --github",
   "repository": {
     "type": "git",
     "url": "https://github.com/readmeio/rdme.git",
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -2,7 +2,7 @@
   "name": "@readme/create",
   "version": "0.0.1",
   "description": "Sync your OpenAPI definition to ReadMe",
-  "bin": "../../bin/rdme openapi --github",
+  "bin": "npx rdme@latest openapi --github",
   "repository": {
     "type": "git",
     "url": "https://github.com/readmeio/rdme.git",
--- a/packages/create/package.json
+++ b/packages/create/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "@readme/create",
+  "version": "0.0.1",
+  "description": "Sync your OpenAPI definition to ReadMe",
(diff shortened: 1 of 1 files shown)


packages/create/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@readme/create",
+  "version": "0.0.1",
+  "description": "Sync your OpenAPI definition to ReadMe",
+  "bin": "../../bin/rdme openapi --github",
```
erunion, line 5: I don't think this will work if you pull `@readme/create` down as it'll try to run `../../bin/rdme` on your machine.

kanadgupta: how is it any different than what's being done in this line? https://github.com/readmeio/rdme/blob/fc22c1833ae7ffa4c05ad470fda1a3d0938a5dfc/package.json#L11

kanadgupta: ohhh wait I think I know what you mean. I bet it's not even installing `rdme` to begin with. duh

erunion: If we publish `@readme/create` to npm we only publish what's in `packages/create`, which won't include the `bin` directory.

erunion: Can we use `npx` here at all?

```
"bin": "npx rdme openapi --github"
```

kanadgupta: jinx lol https://github.com/readmeio/rdme/pull/597/commits/a85690408b07d0faadbbb73036a05ccfdd5bc10d

erunion: Or `@readme/create` can just have `rdme@latest` as a dependency and then we can call that from an `index.js` file like they do in `create-react-app`:

* https://github.com/facebook/create-react-app/blob/main/packages/create-react-app/package.json#L25
* https://github.com/facebook/create-react-app/blob/main/packages/create-react-app/index.js

kanadgupta: the commit I just pushed feels a little lighter than the CRA approach—are there any upsides besides the ability to write internal comments?

erunion: Nah, if the commit you pushed worked I'm cool w/ that.

kanadgupta: the thing is I have no idea how to test it lol

erunion (approved):

PR #599, opened 2022-09-07
fix(docs): don't use beta tag

kanadgupta:
## 🧰 Changes

Noticed our docs are currently using our `8.0.0-beta.1` tag 😬 we have an existing `getPkgVersion` function that I extracted that into its own file and made it a bit fancier so it pulls from the `npm` registry if we need to. Everything else is just related refactors.

## 🧬 QA & Testing

Do tests pass? And does [the docs dry run](https://github.com/readmeio/rdme/runs/8233558870?check_suite_focus=true#step:5:14) successfully insert our `latest` tag (as opposed to `8.0.0-beta.1`) into the docs properly?[^1]

[^1]: _hint: yes_


__tests__/helpers/get-gha-setup.ts
```diff
@@ -33,8 +33,8 @@ export function before(writeFileSyncCb) {
 
   process.env.TEST_CREATEGHA = 'true';
 
-  const spy = jest.spyOn(createGHAObject, 'getPkgVersion');
-  spy.mockReturnValue('7.8.9');
+  const spy = jest.spyOn(getPkgVersion, 'getPkgVersion');
+  spy.mockReturnValue(Promise.resolve('7.8.9'));
```
erunion, line 37: why promiseize this?

kanadgupta: We have to since we're making a fetch call now 😞

erunion: oh i missed that at the bottom of this pr

erunion (approved):

PR #603, opened 2022-09-12
feat(breaking/version): raw versions only, usage cleanup

### LINKED ISSUE #602: Inconsistent usage docs for `versions:create`
A customer wrote in about this re: the rdme `versions:create` command:

> According to [the online documentation](https://github.com/readmeio/rdme/blame/main/README.md#L280), `version` should be provided via an argument called `<version>`, but according to the usage guide this is not the case.

image

kanadgupta:
| 🚥 Fix #602 |
| :-- |

## 🧰 Changes

- [x] **BREAKING** Makes JSON the default output of the `versions` command and removes all logic/deps pertaining to table generation
- [x] **BREAKING** The `versions:update` command now accepts the `version` parameter as the default option (rather than via a `--version` flag) so the usage is consistent with `versions:create` and `versions:delete`
- [x] Cleans up usage docs for all `versions` commands
- [x] Some minor test bed cleanup so we're always clearing out the `configstore` for tests that interact with it
- [x] Exports the `Version` type (i.e. the version object that is `POST`-ed and received from our API) and uses it throughout the codebase where applicable.

## 🧬 QA & Testing

Do the docs changes look good? And do tests pass?


__tests__/cmds/logout.test.ts
```diff
@@ -6,6 +6,10 @@ import configStore from '../../src/lib/configstore';
 const cmd = new Command();
 
 describe('rdme logout', () => {
+  afterEach(() => {
+    configStore.clear();
```
erunion, line 10: Oh is us not clearing this out why sometimes the login test flakes out?

kanadgupta: I think so 😬  the test bed seems a little flakier now that I added so many tests messing with `configstore` so I'm hoping this addresses that

src/lib/baseCommand.ts
```diff
@@ -152,17 +152,19 @@ export default class Command {
       {
         name: 'main',
         type: String,
-        description: 'Should this version be the primary (default) version for your project?',
+        description:
+          "Should this version be the primary (default) version for your project? (Must be 'true' or 'false')",
```
erunion, line 156: Adding so much boilerplate into the help screens for "must be true or false" seems like a lot.

kanadgupta: yeah I don't love this but not sure if it's clear otherwise since the user is just supposed to pass in a string. ideally this should be a boolean but I've given these commands more love than they already deserve lol

erunion (approved):
really not sure why i ever added that table view in but glad to see it hit da bricks

PR #610, opened 2022-09-12
chore(testbed): jest 29, flakiness fixes

kanadgupta:
## 🧰 Changes

This PR bumps Jest and handles a bunch of testbed housekeeping. Not sure if tests are faster[^2], but the flakiness when running them locally should be gone.

- [x] Bumped [`jest`](https://npm.im/jest), [`ts-jest`](https://npm.im/ts-jest), and [`@types/jest`](https://npm.im/@types/jest) to v29[^1] and updated snapshots accordingly (see [their blog post](https://jestjs.io/blog/2022/08/25/jest-29)) 🆙 📸 
- [x] It was always bugging me that we had set our `NODE_ENV` value to `testing` for our testing environment as opposed to the [widely used `NODE_ENV` value of `test`](https://jestjs.io/docs/environment-variables#node_env), so this makes that long overdue change. We were able to make a slight cleanup to our setup file as a result. 🧹 
- [x] [`configstore`](https://npm.im/configstore) has continued to give us grief when running tests locally and I had the idea of including the [`JEST_WORKER_ID`](https://jestjs.io/docs/environment-variables#jest_worker_id) in our `configstore` paths so Jest workers aren't clashing with each other. Now the test bed is editing a few of config files (7 on my local machine) instead of just one. That tiny change completely solved the flakiness on my local machine ✨ 

Sadly, our Jest coverage issues in CI (see #617) continue to persist 😭 

## 🧬 QA & Testing

Do tests pass? And do they _**always**_ pass?

[^2]: I mean... tests do feel faster? But maybe that's just a placebo ¯\_(ツ)_/¯
[^1]: We were struggling to upgrade to Jest 29 for a few weeks now since our `__tests__/tsconfig.json` file wasn't being read properly. After noodling on this for a bit, I discovered that bumping us down to [the legacy `ts-jest` preset](https://kulshekhar.github.io/ts-jest/docs/getting-started/presets) fixed everything :cryparty: I don't love this solution but hopefully it's temporary.


erunion (approved):
glad you were able to finally sort this out

PR #612, opened 2022-09-15
feat(openapi): add `--dryRun` option

darrenyong:
| 🚥 Fix RM-3854 |
| :-- |

## 🧰 Changes

This adds a `--dryRun` option to the `openapi` command so users can debug their spec creation and updating.

## 🧬 QA & Testing


darrenyong:
@kanadgupta, what do you think we should do in the case where a user does a `dryRun` in a directory that only has one file?

we automatically return [this](https://github.com/readmeio/rdme/blob/main/src/lib/prepareOas.ts#L93) notification letting them know that we're attempting to upload it. should we change the copy at all to avoid confusion or is that okay?

kanadgupta:
@darrenyong good question... I think it's okay to leave the copy as is! I think it's clear from these logs what's happening:
```
We found petstore.json and are attempting to upload it.
🎭 dry run! The spec with the id SPEC_ID will be updated
```
I think we should clean up the dry run log copy a bit but I'll leave a suggestion during my actual review!

__tests__/cmds/openapi/index.test.ts
```diff
@@ -328,6 +328,46 @@ describe('rdme openapi', () => {
...
+          workingDirectory: './__tests__/__fixtures__/relative-ref-oas',
+        })
+      ).resolves.toBe('🎭 dry run! The spec at petstore.json will be uploaded to version 1.0.0');
+
+      const output = getCommandOutput();
+      expect(output).toBe(
+        chalk.yellow(
+          `⚠️  Warning! 🎭 dry run! This will not create or update any API definition and is used only for debug purposes!
+
+ℹ️  We found ${spec} and are attempting to upload it.`
```
darrenyong, line 366: @kanadgupta any ideas on how to make this look a little nicer? this expect is kinda ugly right now 😂

__tests__/cmds/openapi/index.test.ts
```diff
@@ -328,6 +328,46 @@ describe('rdme openapi', () => {
...
+        .reply(201, { registryUUID, spec: { openapi: '3.0.0' } });
+
+      const mockWithHeader = getAPIMockWithVersionHeader(version)
+        .get('/api/v1/api-specification')
+        .basicAuth({ user: key })
+        .reply(200, [])
+        .post('/api/v1/api-specification', { registryUUID })
+        .delayConnection(1000)
+        .basicAuth({ user: key })
+        .reply(201, { _id: 1 }, { location: exampleRefLocation });
```
darrenyong, line 348: @kanadgupta I removed the `mock.done()` and `mockWithHeader.done()` at the end of this test since it was making it fail. I'm thinking it was expecting these API calls to be made, but since it's a dry run, it doesn't actually make any api calls from within the creation.

I get ESLint errors with these two since they aren't being called anymore, but removing them also gives a `Received promise rejected instead of resolved` error that I'm not sure how to resolve.

kanadgupta: Hmm you'll need the [`mock.done()`](https://github.com/nock/nock#expectations) statements—that's an important check to ensure that every mock API request has been satisfied. I'd double check that every mocked API request is supposed to be there[^1]. Let me know if you wanna pair on this!

[^1]: hint: because it's a **dry run**, we're definitely not making every API request that's being mocked 👀

__tests__/cmds/openapi/index.test.ts
kanadgupta: Instead of using [`toBe()`](https://jestjs.io/docs/expect#tobevalue) (which is an exact match), you could use [`toMatch()`](https://jestjs.io/docs/expect#tomatchregexp--string) (which is a partial match)! See here for an example: https://github.com/readmeio/rdme/blob/5e4722bc038df93be9f8462b6f080e72e79504e2/__tests__/cmds/openapi/index.test.ts#L588-L589

__tests__/cmds/openapi/index.test.ts
```diff
@@ -483,6 +483,40 @@ describe('rdme openapi', () => {
...
+      const spec = './__tests__/__fixtures__/ref-oas/petstore.json';
+
+      await expect(
+        openapi.run({
+          key,
+          version,
+          spec,
+          dryRun: true,
+        })
+      ).resolves.toMatch('dry run! The spec with the id spec2 will be updated');
```
darrenyong, line 514: you think it's okay to hard code `spec2` here since it's getting hard coded in the `prompts.inject`?

kanadgupta: Yeah I think that's fine! Alternatively you could create a variable called `specId` and use it like this, but I don't think it really matters honestly:
```suggestion
      ).resolves.toMatch(`dry run! The spec with the id ${specId} will be updated`);
```

darrenyong:
@kanadgupta merged main 😎

src/cmds/openapi/index.ts
```diff
@@ -81,13 +82,18 @@ export default class OpenAPICommand extends Command {
         description:
           "Automatically update an existing API definition in ReadMe if it's the only one associated with the current version.",
       },
+      {
+        name: 'dryRun',
+        type: Boolean,
+        description: 'Runs the command without creating/updating any OAS specs in ReadMe. Useful for debugging.',
```
kanadgupta, line 88:
```suggestion
        description: 'Runs the command without creating/updating any API definitions in ReadMe. Useful for debugging.',
```

src/cmds/openapi/index.ts
```diff
@@ -98,6 +104,12 @@ export default class OpenAPICommand extends Command {
      */
     const ignoredGHAParameters: Options = { version: undefined, update: undefined };
 
+    if (dryRun) {
+      Command.warn(
+        '🎭 dry run! This will not create or update any API definition and is used only for debug purposes!'
```
kanadgupta, line 109:
```suggestion
        '🎭 dry run option detected! No API definitions will be created or updated in ReadMe.'
```

src/cmds/openapi/index.ts
```diff
@@ -224,6 +236,10 @@ export default class OpenAPICommand extends Command {
     };
 
     function createSpec() {
+      if (dryRun) {
+        return `🎭 dry run! The spec at ${specPath} will be uploaded to version ${selectedVersion}`;
```
kanadgupta, line 240:
```suggestion
        return `🎭 dry run! The API definition located at ${specPath} will be created for this project version: ${selectedVersion}`;
```

src/cmds/openapi/index.ts
```diff
@@ -237,6 +253,10 @@ export default class OpenAPICommand extends Command {
     }
 
     function updateSpec(specId: string) {
+      if (dryRun) {
+        return `🎭 dry run! The spec with the id ${specId} will be updated`;
```
kanadgupta, line 257:
```suggestion
        return `🎭 dry run! The API definition located at ${specPath} will update this API definition ID: ${specId}`;
```

kanadgupta (changes requested):
A few copy edit suggestions below but the code and tests look great! Could you also add a little note about the dry run flag in the `README.md` file? I think it makes sense for that note to live around [here](https://github.com/readmeio/rdme/blame/ee01e65a8d111896d101d27ec930993150ed799c/README.md#L99).

darrenyong:
> 

Added! I'll update our docs [here](https://docs.readme.com/docs/rdme#dry-run-mode) with the `openapi --dryRun` once we release this!

README.md
```diff
@@ -150,6 +152,8 @@ You can add `--update` to the command so if there's only one API definition for
 rdme openapi [path-to-file.json] --version={project-version} --update
 ```
 
+This command also has a dry run mode, which can be useful for initial setup and debugging. You can read more about dry run mode [in our docs](https://docs.readme.com/docs/rdme#dry-run-mode).
```
kanadgupta, line 155:
```suggestion
This command also has a dry run mode, which can be useful for initial setup and debugging. You can perform a dry run by supplying the `--dryRun` flag.
```

PR #613, opened 2022-09-16
feat(openapi): raw command output option

kanadgupta:
| 🚥 Fix RM-5419 |
| :-- |

## 🧰 Changes

Chatted with a customer who wants to programmatically grab the `id` from the `openapi` command output. Figured this would be a nice win to sneak into v8.

## 🧬 QA & Testing

Any other attributes that you think would be helpful to include in the JSON object? And do tests pass?


erunion (approved):
:nice-rgb:

PR #619, opened 2022-09-28
feat(breaking/login): remove `2fa` flag in favor of better prompts

kanadgupta:
| 🚥 Fix RM-5498 |
| :-- |

## 🧰 Changes

The `--2fa` flag that's required to log in with `rdme` is clunky. Now that we've overhauled our prompts system, it was stupidly easy to add a flow that goes something like this:

1. User logs in with username and password

1. API returns a 401 with the `LOGIN_TWOFACTOR` code, meaning the user forgot to include a 2FA code

1. `rdme` checks for the `LOGIN_TWOFACTOR` code. In that case, it prompts the user for the token and sends another login request with the token![^1]

By removing the `2fa` flag in the `rdme` client, the login flow is infinitely more smooth. The other great thing is that this requires no changes to our API, so we can maintain backwards compatibility with older versions of `rdme`.

This PR also includes some minor testbed housekeeping as well.

## 🧬 QA & Testing

Do these changes make sense to you? And do tests pass?

[^1]: If the API returns any other error, it just throws it as it normally would.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,6 @@ For local CLI usage with a single project, you can authenticate `rdme` to your R
 rdme login
 ```
 
-If you have [two-factor authentication (2FA)](https://docs.readme.com/docs/two-factor-authentication) enabled on your account, you'll need to pass in the `--2fa` option:
-
-```sh
-rdme login --2fa
-```
-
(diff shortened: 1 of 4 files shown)


erunion (commented):
bless you for fixing this.

PR #635, opened 2022-10-12
feat(breaking): drop version creation question from prompt flows

kanadgupta:
## 🧰 Changes

When going through some of the prompts for the `openapi` command, I realize we have the following question in the prompt flow:
```
? Would you like to use an existing project version or create a new one? › - Use arrow-keys. Return to submit.
❯   Use existing
    Create a new version
```
It feels like an unlikely scenario where a user will want to create a new version when running this command. And upon further investigation, we do this for a few other commands for which it makes even less sense to ask this question:

- `categories`: if you're trying to list the categories for a project version, why would you want to create a new version just to list the categories in that newly created version?
- `docs:edit`: if you're trying to edit an existing document, why would you want to create a new version before editing a document within that newly created version?

All in all, the whole "version-creation-within-another-command" feature never appeared to be all that useful and was just creating additional prompt fluff[^3] and unnecessary technical debt[^2], so I ultimately decided it's time we rip it out 

I was able to clean up a lot of code and types in the process. Here's a summary of everything:
- [x] Drastically simplified the version selection prompt logic that's used in like 8 commands and consolidated it in [`src/lib/versionSelect.ts`](https://github.com/readmeio/rdme/pull/635/files#diff-c42c5c43dde763bc34bee8b579a0ef8b355c3bd197a8bde953b84e01265c51d3) so it's no longer a separate function in [`src/lib/prompts.ts`](https://github.com/readmeio/rdme/pull/635/files#diff-198f9c6e4714f4c1819b759cdd66ece2a658ef58c9797c084b86cba0e0224e3f)[^1]
- [x] Consolidated and cleaned up a few `Version` object types
- [x] Updated some tests affected by these two points 👆🏽 

## 🧬 QA & Testing

[^1]: Have I mentioned how I hate this file?
[^2]: I mean look at some of this archaic ternary logic: https://github.com/readmeio/rdme/blob/542307bd9249f6fdad66470564c88a024a6c6166/src/lib/prompts.ts#L44 https://github.com/readmeio/rdme/blob/542307bd9249f6fdad66470564c88a024a6c6166/src/lib/prompts.ts#L57
[^3]: Now that we're using [`prompts`](https://www.npmjs.com/package/prompts) and can rapidly develop prompts more easily, the number of questions we ask when a user runs `rdme openapi` has blown up quite a bit.

### DIFF
--- a/src/lib/prompts.ts
+++ b/src/lib/prompts.ts
@@ -1,3 +1,4 @@
+import type { Version } from '../cmds/versions';
 import type { VersionCreateOptions } from 'cmds/versions/create';
 import type { VersionUpdateOptions } from 'cmds/versions/update';
 import type { Response } from 'node-fetch';
@@ -13,10 +14,6 @@ type SpecList = {
   title: string;
 }[];
 
-type VersionList = {
-  version: string;
-}[];
-
 type ParsedDocs = {
   next?: {
     page: number;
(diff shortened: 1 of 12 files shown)


src/lib/prompts.ts
```diff
@@ -28,41 +25,6 @@ type ParsedDocs = {
...
-      choices: versionList.map(v => {
-        return {
-          title: v.version,
-          value: v.version,
-        };
-      }),
-    },
-    {
-      type: (prev, values) => {
-        return (selectOnly ? true : values.option === 'update') ? null : 'text';
```
erunion, line 57: giphy

erunion (approved):

src/lib/prompts.ts
kanadgupta: this is 🚮 personified, i'm obsessed

PR #636, opened 2022-10-12
feat: prompt for login any time a user omits API key

kanadgupta:
| 🚥 Fix RM-5540 |
| :-- |

## 🧰 Changes

This updates all of our commands so that when they're run locally and the user omits the `--key` parameter (and they're running the command from a non-CI environment), we now prompt the user to log in, rather than just erroring out. Most of this diff is just a super tiny refactor to all commands and then refactoring a bunch of tests to account for this change.

Here's what it looks like when a user omits the `--key` flag and attempts to run a command that requires authentication:

https://user-images.githubusercontent.com/8854718/195464152-916c7330-9cd6-4a17-8db2-177931084099.mov

Also did a tiny bit of test housekeeping:
- Added some nock cleanup logic to our `docs:edit` tests to be consistent with every other test suite 🧹 
- Stricter `login` command nocks 💪🏽 

## 🧬 QA & Testing

Our testbed covers this, but if you want to test this out, you can check out this branch and run this flow of commands:

```sh
# Set up rdme
npm ci && npm run build
# Just in case...
bin/rdme logout
# Now try running an authenticated command and follow the prompts
bin/rdme openapi
```

### DIFF
--- a/__tests__/cmds/categories/index.test.ts
+++ b/__tests__/cmds/categories/index.test.ts
@@ -1,4 +1,5 @@
 import nock from 'nock';
+import prompts from 'prompts';
 
 import CategoriesCommand from '../../../src/cmds/categories';
 import getAPIMock, { getAPIMockWithVersionHeader } from '../../helpers/get-api-mock';
@@ -13,10 +14,19 @@ describe('rdme categories', () => {
 
   afterEach(() => nock.cleanAll());
 
-  it('should error if no api key provided', () => {
-    return expect(categories.run({})).rejects.toStrictEqual(
+  it('should prompt for login if no API key provided', async () => {
+    const consoleInfoSpy = jest.spyOn(console, 'info').mockImplementation();
+    prompts.inject(['this-is-not-an-email', 'password', 'subdomain']);
+    await expect(categories.run({})).rejects.toStrictEqual(new Error('You must provide a valid email address.'));
+    consoleInfoSpy.mockRestore();
+  });
+
+  it('should error in CI if no API key provided', async () => {
+    process.env.TEST_CI = 'true';
+    await expect(categories.run({})).rejects.toStrictEqual(
       new Error('No project API key provided. Please use `--key`.')
     );
+    delete process.env.TEST_CI;
   });
 
   it('should return all categories for a single page', async () => {
--- a/__tests__/cmds/login.test.ts
+++ b/__tests__/cmds/login.test.ts
@@ -36,7 +36,7 @@ describe('rdme login', () => {
   it('should post to /login on the API', async () => {
     prompts.inject([email, password, project]);
 
-    const mock = getAPIMock().post('/api/v1/login').reply(200, { apiKey });
+    const mock = getAPIMock().post('/api/v1/login', { email, password, project }).reply(200, { apiKey });
 
     await expect(cmd.run({})).resolves.toBe('Successfully logged in as user@example.com to the subdomain project.');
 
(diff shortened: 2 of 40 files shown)


src/lib/baseCommand.ts
```diff
@@ -86,13 +88,23 @@ export default class Command {
...
       if (!opts.key) {
-        throw new Error('No project API key provided. Please use `--key`.');
+        if (isCI()) {
+          throw new Error('No project API key provided. Please use `--key`.');
+        } else {
+          // eslint-disable-next-line no-console
+          console.info("Looks like you're missing a ReadMe API key, let's fix that! 🦉");
+          const result = await loginFlow();
+          // eslint-disable-next-line no-console
+          console.info(result);
```
erunion, line 104: left in debug?

src/lib/baseCommand.ts
```diff
@@ -86,13 +88,23 @@ export default class Command {
...
     Command.debug(`opts: ${JSON.stringify(opts)}`);
 
     if (this.args.some(arg => arg.name === 'key')) {
       if (!opts.key) {
-        throw new Error('No project API key provided. Please use `--key`.');
+        if (isCI()) {
+          throw new Error('No project API key provided. Please use `--key`.');
+        } else {
+          // eslint-disable-next-line no-console
+          console.info("Looks like you're missing a ReadMe API key, let's fix that! 🦉");
```
erunion, line 101: Don't we have a console helper somewhere to log out info statements?

src/lib/baseCommand.ts
```diff
@@ -86,13 +88,23 @@ export default class Command {
...
+  async run(opts: CommandOptions<{}>): Promise<string> {
     Command.debug(`command: ${this.command}`);
     Command.debug(`opts: ${JSON.stringify(opts)}`);
 
     if (this.args.some(arg => arg.name === 'key')) {
       if (!opts.key) {
-        throw new Error('No project API key provided. Please use `--key`.');
+        if (isCI()) {
+          throw new Error('No project API key provided. Please use `--key`.');
+        } else {
```
erunion, line 99: Probably don't need an `else` here and I'm surprised ESLint isn't complaining about it.

__tests__/cmds/categories/index.test.ts
```diff
@@ -13,10 +14,19 @@ describe('rdme categories', () => {
 
   afterEach(() => nock.cleanAll());
 
-  it('should error if no api key provided', () => {
-    return expect(categories.run({})).rejects.toStrictEqual(
+  it('should prompt for login if no API key provided', async () => {
```
erunion, line 17: Want to add tests for when the user fully fills out the login form to get an API key? Might be a lot to test/mock out. I'm happy with this as-is.

__tests__/cmds/login.test.ts
```diff
@@ -36,7 +36,7 @@ describe('rdme login', () => {
   it('should post to /login on the API', async () => {
     prompts.inject([email, password, project]);
 
-    const mock = getAPIMock().post('/api/v1/login').reply(200, { apiKey });
+    const mock = getAPIMock().post('/api/v1/login', { email, password, project }).reply(200, { apiKey });
```
erunion, line 39: I really wish there were a way to force nock to error out if you send a payload to a mock that doesn't have one. I wonder if we have any other cases of this around here.

erunion (approved):
well shit, you made this look easy. we should've had these commands do this this whole time.

__tests__/cmds/categories/index.test.ts
kanadgupta: This `loginFlow` function has complete test coverage via the `login` command tests so I'm not gonna worry about that for now.

src/lib/baseCommand.ts
kanadgupta: Oof, nice catch thank you! See https://github.com/readmeio/rdme/pull/636/commits/ce6455da39062952fa55cd86ab4d58221015f894

src/lib/baseCommand.ts
kanadgupta: Good call! Our `info` helper prefixes all statements with a ℹ️ emoji but I extended that helper with a flag to configure that emoji prefix and refactored a bunch of `console.info` statements to use this: https://github.com/readmeio/rdme/commit/bab855db842e095fe6d521acd094607fbc4b33a3

Issue #638, opened 2022-10-15
Question: validate command to report all validation issues instead of one at a time


Ebeid:
Not sure if there is a way for **validate** to report back all validation issues/violations instead of one at a time.

kanadgupta:
Looks to be an upstream issue with how validation errors are returned in [`openapi-parser`](https://github.com/readmeio/openapi-parser). We'll dig into this!

erunion:
This is now resolved in https://github.com/readmeio/rdme/pull/639, https://github.com/readmeio/openapi-parser/pull/135, https://github.com/readmeio/oas-normalize/pull/220 and will be part of out upcoming v8 release.

Ebeid:
@erunion when v8 release scheduled ?

kanadgupta:
@Ebeid most likely in a couple weeks!

kanadgupta:
@Ebeid I've tagged a new beta release if you wish to play around with it.

```sh
npm install rdme@next
```

PR #639, opened 2022-10-17
fix(validate): surface all validation errors

erunion:
| 🚥 Fix https://github.com/readmeio/rdme/issues/638 |
| :-- |

## 🧰 Changes

This updates our [@readme/openapi-parser](https://npm.im/@readme/openapi-parser) and [oas-normalize](https://npm.im/oas-normalize) dependencies to begin surfacing **all** validation errors in OpenAPI and Swagger validation.


package.json
```diff
@@ -50,8 +50,8 @@
     "jsonpath": "^1.1.1",
     "mime-types": "^2.1.35",
     "node-fetch": "^2.6.1",
-    "oas": "^18.4.4",
-    "oas-normalize": "^7.0.0",
+    "oas": "^19.0.1",
```
erunion, line 53: Nothing new there that's going to be used within the library, I just released a v19 last week so might as well update it here.

kanadgupta (approved):
Would you consider this a breaking change? If so, could you rename the PR title from `fix(validate)` to `feat(breaking/validate)`?

erunion:
@kanadgupta No I wouldn't consider this a breaking change.

package.json
kanadgupta: Do we even use `oas` anywhere? Nothing comes up during my quick search

kanadgupta: Ignore me, it's used in the reducer.

erunion: We load it in for the reducer command as that code is in `oas`.

PR #643, opened 2022-10-24
feat(baseCommand): inform user which project they're making changes to

darrenyong:
| 🚥 Fix RM-5582 |
| :-- |

## 🧰 Changes

If users are logged in already, we're going to start informing them which account they're using to make changes and to which project changes are being made to since it currently isn't clear. 

## 🧬 QA & Testing

The info warning should be displayed whenever any commands like `openapi`, `docs`, `changelog` are being used. It shouldn't show up if the user is logged out and there aren't any saved/stored credentials.


darrenyong:
I will definitely need some copy changes 😂

__tests__/index.test.ts
```diff
@@ -117,6 +117,41 @@ describe('cli', () => {
     conf.clear();
   });
 
+  describe('project info', () => {
```
kanadgupta, line 120:
```suggestion
  describe('logged-in user notifications', () => {
```

src/lib/baseCommand.ts
```diff
@@ -93,6 +93,14 @@ export default class Command {
     Command.debug(`opts: ${JSON.stringify(opts)}`);
 
     if (this.args.some(arg => arg.name === 'key')) {
+      if (opts.key && configstore.get('apiKey') === opts.key) {
+        info(
+          `You're currently making updates to the project: ${configstore.get(
+            'project'
+          )} on the account ${configstore.get('email')}`
+        );
```
kanadgupta, line 101: Copy edit[^1] below! Also note I added the `false` since we want to use this key emoji prefix (🔑) rather than ℹ️

```suggestion
        info(
          `🔑 ${configstore.get('email')} is currently logged in, using the stored API key for this project: chalk.blue(${configstore.get(
            'project'
          )})`, false
        );
```

[^1]: you'll definitely have to rerun prettier here, assuming the syntax is even correct lol

__tests__/index.test.ts
```diff
@@ -117,6 +117,41 @@ describe('cli', () => {
...
+      );
+    });
+
+    it('should not inform a logged in user when they pass their own key', async () => {
+      await expect(cli(['docs', '--key=asdf'])).rejects.toThrow(
+        'No folder provided. Usage `rdme docs <folder> [options]`.'
+      );
+
+      expect(getCommandOutput()).toBe('');
+    });
```
kanadgupta, line 152: this is **_so_** clean 🤌

kanadgupta (changes requested):
A few small copy suggestions, but the tests and logic look great 🎉

__tests__/index.test.ts
darrenyong: so DRY 
image

__tests__/index.test.ts
```diff
@@ -108,6 +109,42 @@ describe('cli', () => {
...
+
+      await expect(openapi.run({ key, create: true, update: true })).rejects.toStrictEqual(
+        new Error('The `--create` and `--update` options cannot be used simultaneously. Please use one or the other!')
+      );
+
+      expect(getCommandOutput()).toMatch('is currently logged in, using the stored API key for this project:');
+    });
+
+    it('should not inform a logged in user when they pass their own key', async () => {
+      await expect(cli(['openapi --create --update', '--key=asdf'])).rejects.toThrow('Command not found.');
```
kanadgupta, line 142: That error message doesn't look right 🤔 what happens when you separate out each parameter into separate strings?
```suggestion
      await expect(cli(['openapi', '--create', '--update', '--key=asdf'])).rejects.toThrow('Command not found.');
```

__tests__/index.test.ts
```diff
@@ -108,6 +109,42 @@ describe('cli', () => {
...
+      conf.clear();
+    });
+
+    it('should inform a logged in user which project is being updated', async () => {
+      const openapi = new OpenAPICommand();
+      const key = '123456';
+
+      await expect(openapi.run({ key, create: true, update: true })).rejects.toStrictEqual(
+        new Error('The `--create` and `--update` options cannot be used simultaneously. Please use one or the other!')
+      );
```
kanadgupta, line 136: If my approach in this comment works and the error message is the same as that one, could you use that pattern instead of importing the `openapi.run` function here?

kanadgupta (changes requested):

kanadgupta (approved):
🚀🫡

PR #644, opened 2022-10-26
refactor(docs): move `--cleanup` option into `docs:prune` command

kanadgupta:
| 🚥 Fix RM-5580 |
| :-- |

## 🧰 Changes

Separates the `--cleanup` option on the `docs` command into a separate command called `docs:prune`. The `docs` command logic was getting confusing and it feels like an entirely separate use-case that is worthy of a separate command. This entire command is wrapped in a prompt now so the user must say "yes" before continuing with the pruning.

## 🧬 QA & Testing

Do the changes look good? And do tests pass?


src/cmds/docs/cleanup.ts
```diff
@@ -0,0 +1,102 @@
...
+function getSlug(filename: string): string {
+  const { slug } = readDoc(filename);
+  return slug;
+}
+
+export default class DocsCleanupCommand extends Command {
+  constructor() {
+    super();
+
+    this.command = 'docs:cleanup';
```
erunion, line 29: Thoughts on renaming this to `docs:prune`?

src/cmds/docs/edit.ts
```diff
@@ -30,7 +30,7 @@ export default class EditDocsCommand extends Command {
     this.usage = 'docs:edit <slug> [options]';
```
erunion, line 30: Honestly not opposed to deprecating this command now and removing it in v9.

src/cmds/docs/cleanup.ts
```diff
@@ -0,0 +1,102 @@
...
+        name: 'deleteAll',
+        message: `This action will delete all docs under version ${selectedVersion} from ReadMe, would you like to continue?`,
+      });
+
+      if (!deleteAll) {
+        return Promise.reject(new Error('Aborting, no changes were made.'));
+      }
+    }
+
+    Command.warn(`We're going to delete from ReadMe any document that isn't found in ${folder}.`);
```
erunion, line 88: Should this also have a "are you sure you want to do this?" confirmation prompt? Feels like it might be easy to typo a directory name on this command and wipe all of your docs out without any recourse.

src/cmds/docs/cleanup.ts
kanadgupta: Love it!

src/cmds/docs/cleanup.ts
kanadgupta: Yeah I've sorta felt like we should just wrap this entire command in a confirmation prompt, but I wasn't sure if folks use it in CI. I'll just make that change and if folks request it, we'll create an exception for CI environments.

kanadgupta: Done: https://github.com/readmeio/rdme/pull/644/commits/ac5580717a7aee33f3efe4034c34f17260c73172

src/cmds/docs/cleanup.ts
kanadgupta: Done: https://github.com/readmeio/rdme/pull/644/commits/863d06cdbeedf06a3cc2c6473f42035a04b8609f

kanadgupta:
@shaiarmis FYI!

erunion (approved):

PR #647, opened 2022-10-27
feat(docs): add `guides` alias

kanadgupta:
| 🚥 Fix RM-5581 |
| :-- |

## 🧰 Changes

Adds an alias for the `docs` family of commands called `guides`, as well as some corresponding documentation touch-ups. Note that the `docs:edit` command doesn't have a corresponding alias since that command is now deprecated (see https://github.com/readmeio/rdme/pull/646).

## 🧬 QA & Testing

Does the alias'ing approach + the corresponding docs look good?

### DIFF
--- a/src/lib/commands.ts
+++ b/src/lib/commands.ts
@@ -20,7 +20,7 @@ export function getCategories(): Record<
       commands: [],
     },
     docs: {
-      description: 'Documentation',
+      description: 'Docs (a.k.a. Guides)',
       commands: [],
     },
     changelogs: {
--- a/src/lib/commands.ts
+++ b/src/lib/commands.ts
@@ -16,7 +16,7 @@ export function getCategories(): Record<
 > {
   return {
     apis: {
-      description: 'Upload OpenAPI/Swagger definitions',
+      description: 'OpenAPI / Swagger',
       commands: [],
     },
     docs: {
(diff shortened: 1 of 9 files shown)


src/lib/commands.ts
```diff
@@ -31,14 +31,14 @@ export function getCategories(): Record<
...
     },
-    categories: {
-      description: 'Categories',
-      commands: [],
-    },
     versions: {
       description: 'Versions',
       commands: [],
     },
+    categories: {
```
kanadgupta, line 38: Rearranged this so the categories in the help screen align with the README.md command reference.

src/lib/commands.ts
```diff
@@ -16,11 +16,11 @@ export function getCategories(): Record<
 > {
   return {
     apis: {
-      description: 'Upload OpenAPI/Swagger definitions',
+      description: 'OpenAPI / Swagger',
```
erunion, line 19: wayyyy cleaner of a header, good lordt

erunion (approved):

src/lib/commands.ts
kanadgupta: the amount of little oversights on the **main** help screen that we don't notice for literal years lmao

erunion: 🤡

PR #648, opened 2022-10-31
feat(docs:prune): `--confirm` flag to bypass prompt

shaiarmis:
## 🧰 Changes

Add `--noPrompt` flag to the `docs:prune` command

## 🧬 QA & Testing

1. Manually add a document to your project in the UI.
2. Run the `docs:prune` command as is, see that it asks for confirmation before deleting the document.
3. Run again but with `--noPrompt` and see that the document is deleted from the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ If you wish to delete documents from ReadMe that are no longer present in your l
 rdme docs:prune path-to-directory-of-markdown
 ```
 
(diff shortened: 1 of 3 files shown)


kanadgupta (approved):
Renamed the flag, cleaned up the docs, and did a small refactor to simplify the code a bit. Thanks @shaiarmis!

PR #660, opened 2022-11-03
refactor(createGHA): drop hercule in favor of native solution

kanadgupta:
## 🧰 Changes

As part of our woes with the sindr*sorhus industrial complex, this PR removes `hercule` in favor of a native solution using `String.replace()` that was stupidly clean and simple.

## 🧬 QA & Testing

We have a ton of comprehensive snapshot testing for our GitHub Actions workflow files, so if those tests pass, we should be good to go.


erunion (approved):
🚮

PR #666, opened 2022-11-08
fix: package.json keywords declarations

erunion:
## 🧰 Changes

Apparently `tags` isn't a thing in [the package.json spec](https://docs.npmjs.com/cli/v8/configuring-npm/package-json), which seems to be the reason why our keywords aren't showing up on npm: https://www.npmjs.com/package/rdme

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "bin": {
     "rdme": "bin/rdme"
   },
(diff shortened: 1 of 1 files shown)


kanadgupta (approved):
omfg. thank you

Issue #664, opened 2022-11-07
GHA Onboarding Prompt breaks CI in Jenkins/CloudBees


lanz:
We use rdme to sync our API specs with ReadMe. We have a build job set to run in Jenkins/CloudBees that runs the following command for several specs when content is merged in our repo:

`rdme openapi "output/swagger/spec.json" --key="<redacted>" --id="<redacted>"`

This has worked flawlessly for the past ~2 years; however, after upgrading to the most recent version of the rdme package, we're now being prompted if we would like to add a GitHub action after running the command. This breaks our automated scripts that we use for continuous integration via Jenkins/CloudBees because there is no user to input a response in a continuous integration environment.

I realize that the question is only asked once, and once a user responds with "no" then the question is never asked again. However, as I mentioned, there's no user in a continuous integration environment to input a response at the prompt.

Note that the above command is the recommended approach for continuous integration according to the documentation for the rdme package in npm.

The rdme tool should be changed to not require user input when updating specs using the above command since it breaks automated scripts. At the very least a flag should be added that suppresses user input prompts.

Note that we can't use GitHub Actions for several reasons. Unless the above is fixed we'll be unable to continue updating our specs in ReadMe automatically.

![API](https://user-images.githubusercontent.com/56061103/200404411-13b0528b-ac2e-4bc8-b6b3-d99958b07ccc.png)

kanadgupta:
Hi @lanz! Apologies for the trouble. We use [`ci-detect`](https://github.com/npm/ci-detect) to determine whether or not a user is running via CI (in which case we don't execute that GitHub question flow). Can you confirm that your CI environmental variables are showing up in the list in [`ci-detect`](https://github.com/npm/ci-detect)? I see Jenkins but I'm not sure about your specific configuration.

Also can you confirm that you're running the latest of the latest (v8.0.4)? We've made a couple of small fixes on this front since v8 was released.

lanz:
Hi @kanadgupta, thank you so much for responding! I verified that we're using the latest version, 8.0.4. In the ci-detect file you linked to, I do see Jenkins, but I don't see CloudBees which is an enterprise version variation of Jenkins.

To add some more context, here's the exact failure output from the rdme command in our CloudBees instance:

```
15:34:19  - Validating the API definition located at output/swagger/spec.json...
15:34:19  ✔ Validating the API definition located at output/swagger/spec.json... done! ✅
15:34:19  - Staging your API definition for upload...
[...]

PR #667, opened 2022-11-08
fix: swap out `@npmcli/ci-detect` for `ci-info`

kanadgupta:
## 🧰 Changes

I opened https://github.com/npm/ci-detect/issues/47 to address this, but per #664 it looks like our CI detection has some imperfections. [`ci-info`](https://github.com/watson/ci-info) seems to be more popular and well-maintained. We looked at this package ages ago but I was initially hesitant because it is a bit annoying to write tests with it but I got that figured out.

**Update**: it looks like the npm CLI team [is planning on deprecating their package in favor of `ci-info`](https://github.com/watson/ci-info/pull/95#issue-1444551786), so we should probably do this anyways.

## 🧬 QA & Testing

No material changes here except perhaps better CI detection. Do tests pass?

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "dependencies": {
     "@actions/core": "^1.6.0",
     "chalk": "^4.1.2",
-    "ci-info": "^3.5.0",
+    "ci-info": "^3.6.1",
     "command-line-args": "^5.2.0",
     "command-line-usage": "^6.0.2",
     "config": "^3.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,7 +76,6 @@
     "@types/jsonpath": "^0.2.0",
     "@types/mime-types": "^2.1.1",
     "@types/node-fetch": "^2.6.2",
-    "@types/npmcli__ci-detect": "^2.0.0",
```
erunion, line 79: 🚮

erunion (approved):

PR #668, opened 2022-11-09
ci: extending ci to run tests on windows

erunion:


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# https://prettier.io/docs/en/options.html#end-of-line
+* text=auto eol=lf
```
erunion, line 2: Without this Prettier loses its shit on Windows. 🤷

Screen Shot 2022-11-08 at 4 32 52 PM

erunion:
@kanadgupta Most of these Windows failures seem to be coming from test snapshots having hardcoded path separators in them, and I'm not really sure how to easily fix this, but these failures are very strange:

Screen Shot 2022-11-08 at 5 02 07 PM

I can't find any sort of this regex anywhere, any idea what it's coming from?

kanadgupta:
Ugh. I have a feeling it's related to the `ignore` package we're using:

https://github.com/readmeio/rdme/blob/9d026cb563e466c1e9a9942088b4a613b2bac169/src/lib/readdirRecursive.ts#L24

src/lib/readdirRecursive.ts
```diff
@@ -21,7 +21,7 @@ export default function readdirRecursive(folderToSearch: string, ignoreGit = fal
   let ignoreFilter: Ignore;
   if (ignoreGit) {
     // Initialize ignore filter with `.git` directory
-    ignoreFilter = ignore().add(path.join(folderToSearch, '.git/'));
+    ignoreFilter = ignore().add(path.join(folderToSearch, '.git', path.sep));
```
erunion, line 24: There seems to be a bug in `ignore` where `.git/` doesn't get escaped on Windows.

https://github.com/kaelzhang/node-ignore/issues/81

200710606-07b1cd56-6a97-446e-ac40-7f955e8fadb3

__tests__/helpers/expect-os-agnostic.ts
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable jest/valid-expect */
+const normalizeString = (str: string) => str.replace(/\\+/g, '/');
+
+/**
+ * Normalize all data expectations to be OS-agnostic and match against Unix-based filesystems
+ * that have path separators of a forward slash.
+ *
+ */
+export default function expectOSAgnostic(actual: any) {
```
erunion, line 9: @kanadgupta This was a fun hack at first but the more I deal with these last four failures that don't make sense the more I hate this approach I've done.

..maybe we just don't test Windows

appeardisappear

src/lib/readdirRecursive.ts
```diff
@@ -21,7 +21,18 @@ export default function readdirRecursive(folderToSearch: string, ignoreGit = fal
...
+    let gitDirectory;
+
+    // There's a problem in the `ignore` library where if a path separator if `\`, like it is on
+    // Windows, the library doesn't escape it, resulting in a corrupted regex matcher.
+    if (path.sep === '\\') {
+      gitDirectory = path.join(process.cwd(), '.git');
+    } else {
+      gitDirectory = path.join(process.cwd(), '.git');
+    }
+
```
erunion, line 33:
```suggestion
    let gitDirectory = path.join(process.cwd(), '.git');
```

PR #669, opened 2022-11-09
feat: adding support for uploading and validating Postman collections

erunion:
## 🧰 Changes

With v8 of [oas-normalize](https://www.npmjs.com/package/oas-normalize) we now have the ability to convert Postman collections to OpenAPI definitions thanks to [postman-to-openapi](https://npm.im/postman-to-openapi). This work here integrates that work into our API definition upload and validation flow, which means that folks will now be able to upload Postman collections directly into their ReadMe projects.

😎 🛹


src/lib/prepareOas.ts
```diff
@@ -65,9 +65,18 @@ export default async function prepareOas(path: string, command: 'openapi' | 'ope
           const oas = new OASNormalize(file, { enablePaths: true });
           return oas
             .version()
```
erunion, line 67: I ended up bringing back `.version() in `oas-normalize@8.1.0` (https://github.com/readmeio/oas-normalize/pull/227) but slightly changed the original API it had to give us some more information on the file it's been supplied.

src/lib/prepareOas.ts
```diff
@@ -65,9 +65,18 @@ export default async function prepareOas(path: string, command: 'openapi' | 'ope
...
-            .then(version => {
-              debug(`OpenAPI/Swagger version for ${file}: ${version}`);
-              return version ? file : '';
+            .then(({ specification, version }) => {
+              if (specification === 'openapi') {
+                debug(`OpenAPI version for ${file}: ${version}`);
+              } else if (specification === 'swagger') {
+                debug(`Swagger version for ${file}: ${version}`);
+              } else if (specification === 'postman') {
+                debug(`Postman version for ${file}: ${version}`);
```
erunion, line 74: `version` here will be "unknown" right now because we don't have any way in `oas-normalize` at the moment to know if a collection is v2.0 or v2.1. Maybe this debug message should be something different?

src/lib/prepareOas.ts
```diff
@@ -65,9 +65,18 @@ export default async function prepareOas(path: string, command: 'openapi' | 'ope
...
+              } else {
+                return '';
+              }
+
+              return file;
             })
             .catch(e => {
               debug(`error extracting OpenAPI/Swagger version for ${file}: ${e.message}`);
```
erunion, line 82: Maybe this debug should be more generic now?

```suggestion
              debug(`error extracting API definition specification version for ${file}: ${e.message}`);
```

src/lib/prepareOas.ts
```diff
@@ -65,9 +65,18 @@ export default async function prepareOas(path: string, command: 'openapi' | 'ope
...
+                debug(`OpenAPI version for ${file}: ${version}`);
+              } else if (specification === 'swagger') {
+                debug(`Swagger version for ${file}: ${version}`);
+              } else if (specification === 'postman') {
+                debug(`Postman version for ${file}: ${version}`);
+              } else {
+                return '';
+              }
+
+              return file;
```
kanadgupta, line 79: how about this?
```suggestion
              debug(`specification type for ${file}: ${specification}`);
              debug(`version for ${file}: ${version}`);
              return ['openapi', 'swagger', 'postman'].includes(specification) ? file : '';
```

src/lib/prepareOas.ts
kanadgupta: agree!

kanadgupta (approved):
Small suggestion below but otherwise LGTM!

src/lib/prepareOas.ts
kanadgupta: I know it's a bit hairy to look at the `info.schema` field, but I'd love if we could parse the Postman version out of there (and also use that to determine if it is a postman collection or not). I have yet to see a schema field that isn't `https://schema.getpostman.com/json/collection/v2.1.0/collection.json` or `https://schema.getpostman.com/json/collection/v2.0.0/collection.json`.

PR #671, opened 2022-11-10
feat(open): `rdme open` should have a flag for opening the dash

darrenyong:
| 🚥 Fix RM-5741 |
| :-- |

## 🧰 Changes

We're adding a `--dash` option to `rdme open` so that it'll open the project dashboard in the browser.

## 🧬 QA & Testing


__tests__/cmds/open.test.ts
```diff
@@ -30,4 +34,36 @@ describe('rdme open', () => {
...
   });
+
+  it('should open the dash', () => {
+    expect.assertions(2);
+    configStore.set('project', 'subdomain');
+    configStore.set('apiKey', '12345');
+
+    const version = '1.0';
+    const key = '12345';
+    const versionPayload: Version = {
```
kanadgupta, line 45: :hell-yeah: great usage of the `Version` type here!

__tests__/cmds/open.test.ts
```diff
@@ -13,6 +16,7 @@ describe('rdme open', () => {
 
   it('should error if no project provided', () => {
     configStore.delete('project');
+    configStore.delete('project');
```
kanadgupta, line 19: Don't think you need to do this twice!

```suggestion
```

src/cmds/open.ts
```diff
@@ -21,21 +23,36 @@ export default class OpenCommand extends Command {
     this.cmdCategory = CommandCategories.UTILITIES;
     this.position = 1;
 
-    this.args = [];
+    this.args = [
+      {
+        name: 'dash',
+        type: Boolean,
+        description: 'Opens the current project dashboard.',
```
kanadgupta, line 30: Small copy suggestion:

```suggestion
        description: 'Opens your current ReadMe project dashboard.',
```

src/cmds/open.ts
```diff
@@ -21,21 +23,36 @@ export default class OpenCommand extends Command {
...
     if (!project) {
       return Promise.reject(new Error(`Please login using \`${config.get('cli')} login\`.`));
     }
 
-    const hubURL: string = config.get('hub');
-    const url = hubURL.replace('{project}', project);
+    let url: string;
+
+    if (dash) {
+      const selectedVersion = await getProjectVersion(undefined, configStore.get('apiKey'), true);
```
kanadgupta, line 49: I think we should add a little check here to make sure the `apiKey` config value exists:
```suggestion
      const key = configStore.get('apiKey');
      if (!key) {
        return Promise.reject(new Error(`Please login using \`${config.get('cli')} login\`.`));
      }
      const selectedVersion = await getProjectVersion(undefined, key, true);
```
Could you also add a test for this?

kanadgupta (changes requested):
A couple small suggestions, but this is looking great!

kanadgupta (approved):

__tests__/cmds/open.test.ts
```diff
@@ -30,4 +33,53 @@ describe('rdme open', () => {
...
+
+      function mockOpen(url: string) {
+        expect(url).toBe(dashUrl);
+        return Promise.resolve();
+      }
+
+      return expect(cmd.run({ mockOpen, dash: true, key: '12345' })).resolves.toBe(
+        `Opening ${chalk.green(dashUrl)} in your browser...`
+      );
+      mockRequest.done();
```
erunion, line 67: @kanadgupta @darrenyong This line doesn't get called because it's after a `return`. This might end up leading to some test memory leaks with mocks not getting closed out.

kanadgupta: oops, great catch, thanks! will fix

kanadgupta: wish we had a lint rule for stuff like this

erunion: Yeah I'm surprised that ESLint didn't pick up on it.

PR #681, opened 2022-11-28
feat: proxy support

### LINKED ISSUE #680: proxy configuration
I am trying to upload a json document to our developer portal behind our company proxy.
However, when try to do the upload, I get the following error in our gitlab pipeline:
request to https://dash.readme.com/api/v1/version failed, reason: connect ETIMEDOUT 172.67.215.249:443
 
The timeout seems to happen because of our firewall blocking the connection.
Does the rdme CLI tool support proxy configuration?
Something like 
HTTP_PROXY=xxx
HTTPS_PROXY=xxx

kanadgupta:
| 🚥 Fix #680 |
| :-- |

## 🧰 Changes

Adds the ability to configure a proxy via the `HTTPS_PROXY` environmental variable. As part of this:

- [x] Added a new lint rule to prevent us from using `node-fetch` for `fetch` requests, with an error message nudging the user towards our fetch wrapper
- [x] Docs 

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -1,6 +1,7 @@
 import type { RequestInit, Response } from 'node-fetch';
 
 import mime from 'mime-types';
+// eslint-disable-next-line no-restricted-imports
 import nodeFetch, { Headers } from 'node-fetch';
 
 import pkg from '../../package.json';
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -12,6 +12,16 @@ import { debug } from './logger';
 
 const SUCCESS_NO_CONTENT = 204;
 
+function getProxy() {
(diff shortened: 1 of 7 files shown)


erunion (approved):

src/lib/fetch.ts
```diff
@@ -11,6 +12,16 @@ import { debug } from './logger';
 
 const SUCCESS_NO_CONTENT = 204;
 
+function getProxy() {
+  // this is something of an industry standard env var, hence the checks for different casings
+  const proxy = process.env.HTTPS_PROXY || process.env.https_proxy;
```
erunion, line 17: TIL that env variables aren't case-insensitive

Screen Shot 2022-11-28 at 3 37 36 PM

PR #682, opened 2022-11-29
feat: major version tags

kanadgupta:
## 🧰 Changes

I've always really liked how [`actions/setup-node`](https://github.com/actions/setup-node) and [`actions/checkout` ](https://github.com/actions/checkout) use top-level tags for major versions, which are [recreated every time there's a tag for that major version](https://github.com/actions/setup-node/tags). It makes it a lot easier to point to the latest in a workflow (as well as in the docs):

```yml
steps:
  - uses: actions/checkout@v3
  - uses: actions/setup-node@v3
```

We've been shipping a lot of updates lately (a good thing!), but it's gotta be annoying for people that have Dependabot set up, since they're likely using an outdated version already. Now that we have [a streamlined release process via `npm run bump`](https://github.com/readmeio/rdme/blob/main/MAINTAINERS.md), I figured we should enhance the process a bit so it automatically recreates a major tag with every release.

As part of this:

- [x] I did a bunch of refactoring to try and consolidate our versioning logic as much as possible
- [x] I also [addressed a little deprecation notice with our GitHub Actions outputs](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) (more details on this in the comments)
- [x] This major tag will be surfaced both in our docs and in any workflow files created via our onboarding.

## 🧬 QA & Testing

I've done a few dry runs of `npm run bump` and confirmed that the tag is created:

Screen Shot 2022-11-29 at 9 59 56 AM

Screen Shot 2022-11-29 at 11 30 45 AM

I also tried with a prerelease tag and confirmed that the major tag is **not** created:

Screen Shot 2022-11-29 at 11 22 26 AM

You can follow these steps to try it out yourself:
1. Check out this branch
2. Run something like `npm run bump -- 8.1.2` or `npm run bump -- 8.1.2-beta.0`
3. Inspect the commit created (but do NOT push it) and make sure that everything looks good! Just make sure to delete the tags you create (using `git tag -d [tag]`) and don't push anything to GitHub.


bin/set-major-version-tag.js
```diff
@@ -0,0 +1,33 @@
...
+ * @example v7
+ */
+async function setMajorVersionTag() {
+  const parsedVersion = semverParse(pkg.version);
+
+  if (parsedVersion.prerelease.length) {
+    // eslint-disable-next-line no-console
+    console.warn('Pre-release version, not setting major version tag');
+    return;
+  }
```
kanadgupta, line 20: We could use our `getPkgVersion` and `getMajorPkgVersion` functions here, but I wanted to keep this fairly vanilla so we can refer to it for a hypothetical future blog post. Plus we need some semver logic here that is pretty specific.

__tests__/cmds/docs/__snapshots__/index.test.ts.snap
```diff
@@ -35,7 +35,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Run \`docs\` command 🚀
-        uses: readmeio/rdme@7.8.9
+        uses: readmeio/rdme@v7
```
erunion, line 38: How come all these snapshots are at v7 and not v8?

src/lib/getPkgVersion.ts
```diff
@@ -1,30 +1,54 @@
...
+export async function getPkgVersion(npmDistTag?: npmDistTag): Promise<string> {
   if (npmDistTag) {
     return fetch(registryUrl)
       .then(res => res.json())
       .then(body => body['dist-tags'][npmDistTag])
-      .catch(() => {
+      .catch(err => {
         // eslint-disable-next-line no-console
-        console.error('error fetching version from npm registry');
+        console.error('error fetching version from npm registry:', err.message);
```
erunion, line 41: should this use the logger?

__tests__/cmds/docs/__snapshots__/index.test.ts.snap
kanadgupta: I wanted to make sure the mocked value was different than the actual one:
https://github.com/readmeio/rdme/blob/e7d02ffbeda3c60a69042ac4a417a2c58fb7a968/__tests__/helpers/get-gha-setup.ts#L39

src/lib/getPkgVersion.ts
kanadgupta: done: 1b6b85d9f31299b4b01cfd38612a27f976bd3ab4

PR #696, opened 2022-12-01
ci: caching deps in ci

erunion:


kanadgupta:
It looks like all the Node versions are using the same cache when looking at these test results:

https://github.com/readmeio/rdme/actions/runs/3595384865/jobs/6054805504#step:11:5
https://github.com/readmeio/rdme/actions/runs/3595384865/jobs/6054805622#step:11:5

Is that going to be an issue?

erunion:
Shouldn't be because it's all coming from the same lockfile.

kanadgupta (approved):

PR #698, opened 2022-12-03
feat(openapi): introduction of a new `openapi:inspect` command

erunion:
## 🧰 Changes

This introduces a new <strike>`openapi:uses`</strike> `openapi:inspect` command that does a series of analyses on API definitions across a series of OpenAPI and ReadMe-speciifc datapoints.

Screen Shot 2022-12-07 at 4 41 23 PM

#### OpenAPI
| Datapoint | What we're analyzing |
| :--- | :--- |
| additionalProperties | Does the API definition utilize `additionalProperties`? |
| callbacks | Are there any operation `callbacks`? |
| circularRefs | Are there any circular `$ref` pointers in the spec? Where are they? |
| discriminators | Does the API definition utilize `discriminator` in polymorphic schemas?
| links | Are there any operation `links`? |
| operationTotal | How many total operations are there? |
| mediaTypes | What are all the media types that this API definition accepts or delivers? |
| style | Do any operation parameters have any serialization requirements? |
| polymorphism | Do any schemas utilize polymorphism (`anyof`, `allOf`, and `oneOf`) for composing themselves? |
| securityTypes | What forms of auth does this API support? |
| serverVariables | Any of the server URLs have variables setup? |
| webhooks | Does this API definition use the OpenAPI 3.1 `webhooks` feature for describing webhooks? |
| xml | Do any parameters, request bodies, or responses document XML payloads? |

#### ReadMe

And we also look for usage of ReadMe's OpenAPI extensions:

* `x-default`
* `x-readme.code-samples`
* `x-readme.headers`
* `x-readme.explorer-enabled`
* `x-readme.proxy-enabled`
* `x-readme.samples-languages`

As well as some deprecated extensions that we've got:

* `x-readme.samples-enabled`
* `RAW_BODY` (not technically an extension but a quirk of our Manual API editor)

## 📽️ Demo

https://user-images.githubusercontent.com/33762/205416025-57f25ae2-e1f0-43d9-9728-aaded0a66450.mov

## 🧬 QA & Testing

Easier to test this locally with a full build instead of using `ts-node` so run `npm run build` and then these commands:

### Full report

```
./bin/rdme openapi:inspect https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/schema-circular.json
```

Screen Shot 2022-12-07 at 4 38 07 PM

### Feature report

> **Note**
>
> Note that the circular refs in this schema are being surfaced. 👀

```
./bin/rdme openapi:inspect https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/schema-circular.json --feature=additionalProperties --feature=style --feature=circularRefs --feature=readme
```

Screen Shot 2022-12-07 at 4 38 36 PM

### Feature report (fancy syntax) 

```
./bin/rdme openapi:inspect https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/schema-circular.json --feature=additionalProperties --feature={style,circularRefs,readme}
```

Screen Shot 2022-12-07 at 4 39 15 PM


src/cmds/openapi/uses.ts
```diff
@@ -0,0 +1,150 @@
...
+export type Options = {
+  spec?: string;
+  workingDirectory?: string;
+};
+
+export default class OpenAPIUsesCommand extends Command {
+  constructor() {
+    super();
+
+    this.command = 'openapi:uses';
```
erunion, line 22: I'm not sold on `:uses` for this command name and still wonder if `:inspect` or `:analyze` are better suited for what all it does.

src/cmds/openapi/index.ts
```diff
@@ -34,7 +34,7 @@ export default class OpenAPICommand extends Command {
     super();
 
     this.command = 'openapi';
-    this.usage = 'openapi [file] [options]';
+    this.usage = 'openapi [file|url] [options]';
```
erunion, line 37: [oas-normalize](https://npm.im/oas-normalize), which all of these commands use for spec processing, handles URLs so we should document in these commands that you can supply them either of these.

kanadgupta:
Just played around, really loving it!

> We talked earlier today about having it return a process.exit(1) exit code if the supplied feature(s) didn't exist in the spec but I'm not sure if we can actually write tests for that without killing Jest.

What if [this function](https://github.com/readmeio/rdme/blob/900ded0e03a1129d1474cb3ad326cef91c0485b0/src/cmds/openapi/uses.ts#L84-L108) returned a [`Promise.allSettled` result](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) (where each result is fulfilled if it contains the feature and rejected if not), and then [this line](https://github.com/readmeio/rdme/blob/900ded0e03a1129d1474cb3ad326cef91c0485b0/src/cmds/openapi/uses.ts#L224) rejects the promise if any of the results contain `status === 'rejected'`?

> I'm not really sure how to best let people supply --feature args for our ReadMe extension features in the analysis report as they have pretty unwieldy names like x-readme.headers or x-readme.explorer-enabled. Doing --feature=x-readme.headers seems pretty shitty. Maybe there should just be a catchall "show all ReadMe extension usage" argument instead and let --feature be OpenAPI-specific?

Yeah I like this "show all extension usage" approach! Our extensions may change over time too so I don't really think those are conventions our users should need to know. Now that I'm thinking about it[^1], it'd be cool if we had our extensions documented in the TS types in `@readme/oas-extensions` and then we were able to pull those into here.

> Also the circularRefs report isn't totally accurate right now as it doesn't tell you which of those refs is circular, it just dumps out all $ref pointers that we didn't dereference. I have some ideas on how to improve this but those changes will happen upstream in oas when I get to that point.

👍🏽

[^1]: doesn't have to happen in this PR, obviously

__tests__/cmds/openapi/__snapshots__/uses.test.ts.snap
```diff
@@ -0,0 +1,184 @@
...
+ · #/components/schemas/FormDataBodyPart/properties/parameterizedHeaders/additionalProperties
+ · #/components/schemas/FormDataContentDisposition/properties/parameters/additionalProperties
+ · #/components/schemas/MultiPart/properties/headers/additionalProperties
+ · #/components/schemas/MultiPart/properties/mediaType/properties/parameters/additionalProperties
+ · #/components/schemas/MultiPart/properties/parameterizedHeaders/additionalProperties
+ · #/components/schemas/ParameterizedHeader/properties/parameters/additionalProperties
+
+circularRefs:
+ · #/components/schemas/MultiPart/properties/parent
+ · #/components/schemas/ZoneOffset/properties/rules"
```
erunion, line 45: 👀

__tests__/cmds/openapi/uses.test.ts
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable jest/no-conditional-expect */
```
erunion, line 1: `eslint-plugin-jest` doesn't like me doing `expect()` calls inside a `try-catch` and it wants me to instead write a wrapper for the command call and no thanks I'd rather not.

https://github.com/jest-community/eslint-plugin-jest/blob/v27.1.6/docs/rules/no-conditional-expect.md#how-to-catch-a-thrown-error-for-testing-without-violating-this-rule

package.json
```diff
@@ -47,17 +47,19 @@
     "form-data": "^4.0.0",
     "gray-matter": "^4.0.1",
     "ignore": "^5.2.0",
-    "jsonpath": "^1.1.1",
+    "jsonpath-plus": "^7.2.0",
```
erunion, line 50: I swapped this out because I'm using `jsonpath-plus` in `oas` for this new `analyzer` library and it doesn't make sense to have rdme ship both in users' `node_modules/` directories.

Why did I use `jsonpath-plus` instead of [jsonpath](https://www.npmjs.com/package/jsonpath)? [jsonpath-plus](https://www.npmjs.com/package/jsonpath-plus) has a nicer API than, and a larger install base and has been recently updated within the past few months vs 2 years.

src/cmds/openapi/uses.ts
```diff
@@ -0,0 +1,272 @@
...
+    });
+
+    spinner.succeed(`${spinner.text} done! ✅`);
+
+    if (features) {
+      const { report, hasUnusedFeature } = this.buildFeaturesReport(analysis, features);
+      if (hasUnusedFeature) {
+        // If we have any unused features we should reject the command with a soft error so we
+        // output the report as normal but return a `exit(1)` status code.
+        return Promise.reject(new SoftError(report));
```
erunion, line 264: We talked about doing this as a `Promise.allSettled` but it was much easier to have `this.buildFeaturesReport` return a boolean for this instead of overhauling the already complex report composition to fit into the structure of a Promise.

src/lib/analyzeOas.ts
```diff
@@ -0,0 +1,213 @@
...
+     * RAW_BODY is specific to our Manual API editor and we don't recommend anyone writing their
+     * own API definition should use it so this is considered deprecated.
+     */
+    raw_body?: AnalyzedFeature;
+  };
+}
+
+const OPENAPI_FEATURE_DOCS: Record<keyof Analysis['openapi'], Pick<AnalyzedFeature, 'description' | 'url'>> = {
+  additionalProperties: {
+    description: 'additionalProperties allows you to document dictionaries where the keys are user-supplied strings.',
```
erunion, line 55: All of these descriptions could definitely use some work I'm sure.

src/lib/analyzeOas.ts
```diff
@@ -0,0 +1,213 @@
...
+    return analysis;
+  });
+}
+
+export function getSupportedFeatures() {
+  return [
+    // OpenAPI features
+    ...Object.keys(OPENAPI_FEATURE_DOCS),
+
+    'readme', // A catch-all for ReadMe features and extensions. Will look for everything.
```
erunion, line 209: Because having people do `--feature=x-readme.explorer-enable` is gnar they can do `--feature=readme` and we'll just query all of their ReadMe extension and feature usage instead. They won't be able to do piecemeal lookups on these but they get a cleaner API instead looking for some of the data they want.

src/lib/prepareOas.ts
```diff
@@ -54,13 +65,18 @@ export default async function prepareOas(path: string, command: 'openapi' | 'ope
...
-    let action: 'upload' | 'reduce' | 'validate';
+    let action: 'inspect' | 'reduce' | 'upload' | 'validate';
     switch (command) {
       case 'openapi':
         action = 'upload';
         break;
+      case 'openapi:uses':
+        // With the messaging that this utility generates, referring to this action as `inspect`
+        // and not `uses` makes more sense.
+        action = 'inspect';
```
erunion, line 76: This is in regard to this messaging:

Screen Shot 2022-12-07 at 4 55 59 PM

erunion:
@kanadgupta @darrenyong @Dashron I found a bunch of small UI bugs last night while looking again at the screenshots I made which just pushed up a bunch of fixes for so this is finally ready for review now.

src/lib/prepareOas.ts
kanadgupta: yeaaah another reason why I wouldn't hate `inspect` as the command name.

src/cmds/openapi/uses.ts
```diff
@@ -0,0 +1,282 @@
...
+        msg = `You have a single ${info.name} in your API.`;
+      }
+
+      report.push(` · ${msg}`);
+    });
+
+    // Build out a view of all OpenAPI and ReadMe features that we discovered.
+    [
+      { component: 'openapi', header: 'OpenAPI Features' },
+      { component: 'readme', header: 'ReadMe-specific Features and Extensions' },
```
kanadgupta, line 185: Per AP-style title casing:
```suggestion
      { component: 'readme', header: 'ReadMe-Specific Features and Extensions' },
```

src/cmds/openapi/uses.ts
```diff
@@ -0,0 +1,282 @@
...
+      spinner.start('Analyzing your API definition for OpenAPI and ReadMe feature usage...');
+    }
+
+    const analysis = await analyzeOas(parsedBundledSpec).catch(err => {
+      Command.debug(`analyzer err: ${err.message}`);
+      spinner.fail();
+      throw err;
+    });
+
+    spinner.succeed(`${spinner.text} done! ✅`);
```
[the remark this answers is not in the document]
kanadgupta: actually on second thought, I like the spinner for individual features and think it's fine to keep it in those cases but still think it should be removed for the entire report!

kanadgupta (changes requested):
mostly small nits, but this feels great when playing around locally and I'm super excited 😌

src/lib/analyzeOas.ts
```diff
@@ -0,0 +1,213 @@
...
+     * This extension is deprecated.
+     */
+    'x-readme.samples-enabled'?: AnalyzedFeature;
+
+    /**
+     * RAW_BODY is specific to our Manual API editor and we don't recommend anyone writing their
+     * own API definition should use it so this is considered deprecated.
+     */
+    raw_body?: AnalyzedFeature;
+  };
```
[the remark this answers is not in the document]
erunion: Reason I did it this way is because I have this type in `oas` because that's where all these queries exist and I'm augmenting it to be `AnalyzedFeature` because I didn't want to store the `description` and `url` information for each of these over in that repo and it didn't feel right living there.

src/lib/prepareOas.ts
```diff
@@ -149,8 +165,10 @@ export default async function prepareOas(path: string, command: 'openapi' | 'ope
     return capitalizeSpecType(type);
   });
 
+  const definitionVersion = await oas.version();
+
   // If we were supplied a Postman collection this will **always** convert it to OpenAPI 3.0.
-  const api = await oas.validate({ convertToLatest: false }).catch((err: Error) => {
+  const api = await oas.validate({ convertToLatest: opts.convertToLatest }).catch((err: Error) => {
```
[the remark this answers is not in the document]
erunion: I'm going to leave this as-is because TS isn't picking up errors when options not available on `oas.validate()` are supplied here:

Screen Shot 2022-12-08 at 2 33 34 PM

Screen Shot 2022-12-08 at 2 32 48 PM

Screen Shot 2022-12-08 at 2 32 54 PM

src/lib/softError.ts
```diff
@@ -0,0 +1,9 @@
+export default class SoftError extends Error {
+  code: string;
+
```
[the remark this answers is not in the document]
erunion: holdover from `APIError`

__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -30,8 +30,10 @@ Options
 
 Related commands
 
-  $ rdme openapi:reduce     Reduce an OpenAPI definition into a smaller subset. 
-  $ rdme openapi:validate   Validate your OpenAPI/Swagger definition.           
+  $ rdme openapi:reduce     Reduce an OpenAPI definition into a smaller subset.                     
+  $ rdme openapi:validate   Validate your OpenAPI/Swagger definition.                               
+  $ rdme openapi:uses       Analyze an OpenAPI/Swagger definition for various OpenAPI and ReadMe
```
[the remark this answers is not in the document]
erunion: talked offline and we're going to dump the ordering system and default to alphabetizing commands in their respective groups. i'll do this in a follow up pr

src/cmds/openapi/uses.ts
erunion: giphy

src/cmds/openapi/uses.ts
erunion: talked offline and we're going to go with `:inspect`

kanadgupta (approved):
🥳

Issue #617, opened 2022-09-20
Jest coverage is broken in GitHub Actions


kanadgupta:
Been seeing this error when running tests in GitHub Actions CI for a bit now:

```
        Failed to write coverage reports:
        ERROR: Error: ENOENT: no such file or directory, open '/home/runner/work/rdme/rdme/coverage/coverage-final.json'
        STACK: Error: ENOENT: no such file or directory, open '/home/runner/work/rdme/rdme/coverage/coverage-final.json'
    at Object.openSync (fs.js:498:3)
    at FileWriter.writeFile (/home/runner/work/rdme/rdme/node_modules/istanbul-lib-report/lib/file-writer.js:185:41)
    at JsonReport.onStart (/home/runner/work/rdme/rdme/node_modules/istanbul-reports/lib/json/index.js:17:45)
    at Visitor.value (/home/runner/work/rdme/rdme/node_modules/istanbul-lib-report/lib/tree.js:38:38)
    at ReportTree.visit (/home/runner/work/rdme/rdme/node_modules/istanbul-lib-report/lib/tree.js:126:17)
    at JsonReport.execute (/home/runner/work/rdme/rdme/node_modules/istanbul-lib-report/lib/report-base.js:12:44)
    at /home/runner/work/rdme/rdme/node_modules/@jest/reporters/build/CoverageReporter.js:241:12
    at Array.forEach (<anonymous>)
    at CoverageReporter.onRunComplete (/home/runner/work/rdme/rdme/node_modules/@jest/reporters/build/CoverageReporter.js:229:25)
    at ReporterDispatcher.onRunComplete (/home/runner/work/rdme/rdme/node_modules/@jest/core/build/ReporterDispatcher.js:71:9)
```

Not sure why, but jest coverage started breaking somewhere between [this commit](https://github.com/readmeio/rdme/pull/577/commits/ca998bc2fae30319a7d85ced85e0f83494fcd221) and [this commit](https://github.com/readmeio/rdme/pull/577/commits/65a209a40981e55db0c861ce53aafbd290501f84).

Anything weird jumping out here? 🧐 https://github.com/readmeio/rdme/compare/ca998bc2fae30319a7d85ced85e0f83494fcd221...65a209a40981e55db0c861ce53aafbd290501f84

See #701 #611 #610 for various attempts to address this.

erunion:
Weird that `rdme` is in there twice:

```
/home/runner/work/rdme/rdme/coverage/coverage-final.json
```

[...]

PR #700, opened 2022-12-05
test: cleaning up our jest config

erunion:
## 🧰 Changes

While upgrading `oas` to Jest 29 I also encountered the same issue we had here where `ts-jest` wasn't picking up our `tsconfig.json` file. Turns out that because we're setting up `transform` on `.ts` files ourselves we don't need this extra `presets` config as it's already being loaded from the `transform` config. Loading both of these was causing the first preset to get loaded without unit test `tsconfig`.

Kind of an annoying gotcha but all tests should still be passing with this change.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,13 +12,6 @@ module.exports = {
   },
   modulePaths: ['<rootDir>'],
   modulePathIgnorePatterns: ['<rootDir>/dist/'],
-  /**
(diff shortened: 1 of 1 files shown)


kanadgupta (approved):
weird, I could have sworn I tried it without a `preset` value in #610 but maybe not. maybe `ts-jest` finally got their shit together?

kanadgupta:
also very bummed that #617 is still happening 😭

PR #707, opened 2022-12-08
feat: dropping our `position` system for ordering commands in `--help`

erunion:
## 🧰 Changes

We've long had this `position` system within our individual commands for positioning where they show up on `--help` screens with the goal of prioritizing the more important commands over those that aren't, however we've always just named our important commands to not have an `:action` in their name so they've always been at the top of the list anyways.

The only `--help` change that's happening with this removal is moving `versions:delete` from being at the end of its group to now being after `versions:create` (bfd):

screen_shot_2022-12-08_at_2 26 17_pm_360

### Main `--help`

Screen Shot 2022-12-08 at 3 40 26 PM

### Command `--help`
Screen Shot 2022-12-08 at 3 39 56 PM


src/lib/help.ts
```diff
@@ -7,9 +7,9 @@ import config from 'config';
 
 import * as commands from './commands';
 
-function formatCommands(cmds: { name: string; description: string; hidden: boolean; position: number }[]) {
+function formatCommands(cmds: { name: string; description: string; hidden: boolean }[]) {
   return cmds
-    .sort((a, b) => (a.position > b.position ? 1 : -1))
+    .sort((a, b) => (a.name > b.name ? 1 : -1))
```
kanadgupta, line 12: a lot of our behind-the-scenes logic here is such a mess lol. the field is called `command` in the command file but `name` here. got it.

kanadgupta (approved):

PR #713, opened 2022-12-16
fix(`docs:prune`): prune child docs

### LINKED ISSUE #710: Prune command doesn't remove child pages (and their parent pages)
When pruning docs, the relevant block of code queries `getCategoryDocs` which returns top level pages, but not the child pages inside that category, meaning the command doesn't cascade out and detect child pages for deletion.

## Repro

1. Create a category with a page, and a child page of that page.
2. Sync it to Readme
3. Delete the files from your local filesystem
4. run `rdme docs:prune [path-to-directory-of-markdown]`
5. See that childPage is still on readme, as is the parent doc

(step 5 is kind of expected because you can't have a hanging page like that, but the CLI doesn't even know to delete the child page first)

https://github.com/readmeio/rdme/blob/d1f6c03636d5907afb5584f416fe666c08c7c06d/src/lib/getDocs.ts#L40-L46

kanadgupta:
| 🚥 Fixes #710 |
| :---------------- |

## 🧰 Changes

Turns out our response payloads from [`GET /categories/{slug}/docs`](https://docs.readme.com/main/reference/getcategorydocs) looks like this for [pages that contain subpages](https://docs.readme.com/main/docs/navigating-your-hub):

```json
[
  ...
  {
    "title": "Navigating Your Hub",
    "slug": "navigating-your-hub",
    "order": 4,
    "hidden": false,
    "children": [
      {
        "title": "Search + Search API",
        "slug": "search",
        "order": 0,
        "hidden": false
      },
      {
        "title": "Quick Nav in the API Reference",
        "slug": "quick-nav-api-reference",
        "order": 1,
        "hidden": false
      }
    ]
  }
]
```

This PR fixes our `docs:prune` command to account for child pages.

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/__tests__/cmds/docs/prune.test.ts
+++ b/__tests__/cmds/docs/prune.test.ts
@@ -122,6 +122,42 @@ describe('rdme docs:prune', () => {
     versionMock.done();
   });
 
+  it('should delete doc and its child if they are missing', async () => {
+    prompts.inject([true]);
+
+    const versionMock = getAPIMock().get(`/api/v1/version/${version}`).basicAuth({ user: key }).reply(200, { version });
+
+    const apiMocks = getAPIMockWithVersionHeader(version)
+      .get('/api/v1/categories?perPage=20&page=1')
+      .basicAuth({ user: key })
+      .reply(200, [{ slug: 'category1', type: 'guide' }], { 'x-total-count': '1' })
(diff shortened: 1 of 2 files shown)


erunion (approved):
api v2 can't come soon enough

kanadgupta:
@erunion 💯

PR #721, opened 2022-12-20
feat(api): surface `Warning` response headers

kanadgupta:
## 🧰 Changes

As we start surfacing [`Warning` headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Warning) from the ReadMe API, we can now surface them in our formatted warning output.

## 🧬 QA & Testing

Don't have an example API request but do tests pass?


erunion (approved):
:hellyea: :nice: :nice-rgb: :sick:

PR #729, opened 2023-01-04
feat(docs): skip markdown files that lack attributes

### LINKED ISSUE #539: Sync files which only has `title` and `category` attributes in the markdown files
It could be interesting to have a configuration flag to only sync Markdown files which match the requirements to be uploaded. 

*Why is this useful?* In your example you have a  `documentation` directory, but in some repositories, the docs are spread on different directories. Instead of running the CLI command multiple times indicating those directories, it could be interesting to indicate the root directory, and mark the flag like `--silent-continue` instead of throw an error because `README.md` doesn't have the attribute `category`.

What do you think about It?

kanadgupta:
| 🚥 Fixes #539 |
| :---------------- |

## 🧰 Changes

This updates our `docs`/`changelogs`/`custompages` commands to skip over any Markdown files that lack YAML frontmatter attributes.

Of all of the requests we get for docs syncing improvements (more are in the pipeline, we promise!), this one is by far the lowest hanging fruit and doesn't require a breaking change[^1], in my opinion.

## 🧬 QA & Testing

I updated our test bed to reflect this, but you can test this out yourself by checking out this branch and running the following command:

```sh
npm run debug -- changelogs README.md
```

[^1]: The reason I don't think it's a breaking change is because we require a `title` attribute on every page type, so if a Markdown file lacked frontmatter attributes, `rdme` would almost certainly error out on it. With this change, `rdme` will skip over it instead and proceed with everything else as normal.


documentation/rdme.md
```diff
@@ -67,7 +67,20 @@ category: 62056dee230e07007218be06
...
+| Attribute  | Required for `changelogs`? | Required for `custompages`? | Required for `docs`? |
+| :--------- | :------------------------- | :-------------------------- | :------------------- |
+| `title`    | Yes                        | Yes                         | Yes                  |
+| `category` | No                         | No                          | Yes                  |
+
+To determine what your `category` value should be, you can use [the `Get all categories` endpoint](https://docs.readme.com/reference/getcategories) and grab the `id` value from the response.
+
+> 📘
+>
+> Any Markdown/HTML files that lack YAML frontmatter attributes will be skipped.
```
erunion:
```suggestion
> Any Markdown/HTML files that lack YAML [frontmatter](https://npm.im/frontmatter) attributes will be skipped.
```

erunion (approved):

documentation/rdme.md
kanadgupta: We already have a front matter link here: https://github.com/readmeio/rdme/blame/6185e5f7d69b9d25509a58a99af8ae449ecdfe61/documentation/rdme.md#L58

PR #733, opened 2023-01-25
ci: use var instead of secret

kanadgupta:
## 🧰 Changes

GitHub Actions now supports variables, so swapping out the version secret for a version variable.

Further reading: https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/

## 🧬 QA & Testing

In the [docs syncing dry run step](https://github.com/readmeio/rdme/actions/runs/4008520309/jobs/6882778511#step:8:10), note how the version is no longer masked:

CleanShot 2023-01-25 at 11 48 03@2x

Here's what it looks like for [the same workflow in `main`](https://github.com/readmeio/rdme/actions/runs/3877892669/jobs/6613439893#step:8:10):

CleanShot 2023-01-25 at 11 48 56@2x

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Sync docs to ReadMe (dry run)
         uses: ./
         with:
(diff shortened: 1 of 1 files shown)


erunion (approved):
nice

PR #735, opened 2023-01-26
feat(GHA): add req header containing file URL

kanadgupta:
| 🚥 Fixes RM-6466 |
| :---------------- |

## 🧰 Changes

**tl;dr** we're adding two new request headers when hitting the ReadMe API! This will facilitate some upcoming changes (see RM-5907) 👀

Here are the two new request headers:

1. **`x-rdme-ci`**: if the command is being run in a CI environment, this will be set to the name of the CI environment as determined by [`ci-info`](https://github.com/watson/ci-info) (e.g., `GitHub Actions`[^1])
1. **`x-readme-source-url`**: if the command is syncing an OpenAPI or Markdown doc to ReadMe, this header[^3] contains the source URL for the file (if applicable). There are two ways that we determine the source file:
    1. If the command is being run via GitHub Actions and a local file is being synced, we use a variety of [default environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) to construct a full URL to the file that looks something like [this](https://github.com/readmeio/rdme/blob/cb4129d5c7b51ff3b50f933a9c7d0c3d0d33d62c/documentation/rdme.md).
    2. If the `openapi` command is being run and the file in question is located at an external URL[^2], that URL is used.

As expected, a bunch of other stuff was done in this PR to facilitate these changes:
- [x] I added a new restricted import ESLint rule for [`ci-info`](https://npm.im/ci-info) — it continues to be very annoying to write tests with and we should confine all usage of the package to our helper functions in `src/lib/isCI.ts`.
- [x] A buttload of tests (including a new test helper and rearranging a few existing tests) 🍑
- [x] Created a few types derived from `oas-normalize` (see https://github.com/readmeio/oas-normalize/pull/248)
- [x] Enhanced our debug logs to include sanitized request headers for API requests
- [x] A few JSDoc additions/cleanups and variable renaming

## 🧬 QA & Testing

Well since each of these CI runs hit the ReadMe API to update [this test project's reference docs](https://rdme-test.readme.io), I was able to go into our Metrics dashboard for the ReadMe API and confirm that the request headers are showing up as expected:

CleanShot 2023-01-31 at 19 07 20@2x

If you go to the API Calls section in the dashboard, filter for my email address, and look for the `rdme-test` in the **user** column, you can click into any one of the API logs and peep the request headers. Note that the URL is off because we rely on [unconventional checkout pathing logic in this workflow](https://github.com/readmeio/rdme/blob/e3c7f22352d6f7e3d24c7fa762df788c905f25e4/.github/workflows/ci.yml#L42-L51), but I feel good about these changes with the test coverage.

[^1]: See the full list here: https://github.com/watson/ci-info#supported-ci-tools (the **Name** column is what will be used)

[^2]: For example, `rdme openapi https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore.json`

[^3]: For commands like `docs` and `openapi`, we do a lot of GET requests as part of the general workflow, but this request header will only be added to the `POST`/`PUT` requests that directly update Markdown/OpenAPI files.

### DIFF
--- a/__tests__/helpers/setup-gha-env.ts
+++ b/__tests__/helpers/setup-gha-env.ts
@@ -1,0 +1,40 @@
+import * as isCI from '../../src/lib/isCI';
+
+/**
+ * A helper function for setting up tests for simulating a GitHub Actions runner environment
+ */
+export function before() {
+  // List of all GitHub Actions env variables:
+  // https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+  process.env.GITHUB_ACTION = '__repo-owner_name-of-action-repo';
+  process.env.GITHUB_ACTIONS = 'true';
+  process.env.GITHUB_REPOSITORY = 'octocat/Hello-World';
+  process.env.GITHUB_RUN_ATTEMPT = '3';
+  process.env.GITHUB_RUN_ID = '1658821493';
+  process.env.GITHUB_RUN_NUMBER = '3';
+  process.env.GITHUB_SERVER_URL = 'https://github.com';
+  process.env.GITHUB_SHA = 'ffac537e6cbbf934b08745a378932722df287a53';
+  process.env.TEST_RDME_CI = 'true';
+
+  const isGHASpy = jest.spyOn(isCI, 'isGHA');
+  isGHASpy.mockReturnValue(true);
+
+  const ciNameSpy = jest.spyOn(isCI, 'ciName');
+  ciNameSpy.mockReturnValue('GitHub Actions (test)');
+}
+
+/**
+ * A helper function for tearing down tests after simulating a GitHub Actions runner environment
+ */
+export function after() {
+  delete process.env.GITHUB_ACTION;
+  delete process.env.GITHUB_ACTIONS;
+  delete process.env.GITHUB_REPOSITORY;
+  delete process.env.GITHUB_RUN_ATTEMPT;
+  delete process.env.GITHUB_RUN_ID;
+  delete process.env.GITHUB_RUN_NUMBER;
+  delete process.env.GITHUB_SERVER_URL;
+  delete process.env.GITHUB_SHA;
+  delete process.env.TEST_RDME_CI;
+  jest.resetAllMocks();
+}
--- a/__tests__/helpers/setup-gha-env.ts
+++ b/__tests__/helpers/setup-gha-env.ts
@@ -1,9 +1,12 @@
 import * as isCI from '../../src/lib/isCI';
 
+const env = process.env;
+
 /**
  * A helper function for setting up tests for simulating a GitHub Actions runner environment
  */
 export function before() {
+  jest.resetModules();
   // List of all GitHub Actions env variables:
   // https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
   process.env.GITHUB_ACTION = '__repo-owner_name-of-action-repo';
@@ -27,14 +30,7 @@ export function before() {
  * A helper function for tearing down tests after simulating a GitHub Actions runner environment
  */
 export function after() {
-  delete process.env.GITHUB_ACTION;
-  delete process.env.GITHUB_ACTIONS;
-  delete process.env.GITHUB_REPOSITORY;
-  delete process.env.GITHUB_RUN_ATTEMPT;
-  delete process.env.GITHUB_RUN_ID;
-  delete process.env.GITHUB_RUN_NUMBER;
-  delete process.env.GITHUB_SERVER_URL;
-  delete process.env.GITHUB_SHA;
+  process.env = env;
   delete process.env.TEST_RDME_CI;
   jest.resetAllMocks();
 }
--- a/__tests__/lib/fetch.test.ts
+++ b/__tests__/lib/fetch.test.ts
@@ -4,46 +4,14 @@ import { Headers } from 'node-fetch';
 
 import pkg from '../../package.json';
 import fetch, { cleanHeaders, handleRes } from '../../src/lib/fetch';
-import * as isCI from '../../src/lib/isCI';
 import getAPIMock from '../helpers/get-api-mock';
+import { after, before } from '../helpers/setup-gha-env';
 
 describe('#fetch()', () => {
   describe('GitHub Actions environment', () => {
-    let isGHASpy: jest.SpyInstance;
-    let ciNameSpy: jest.SpyInstance;
+    beforeEach(() => before());
 
-    // List of all GitHub Actions env variables:
-    // https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
-    beforeEach(() => {
-      process.env.GITHUB_ACTION = '__repo-owner_name-of-action-repo';
-      process.env.GITHUB_ACTIONS = 'true';
-      process.env.GITHUB_REPOSITORY = 'octocat/Hello-World';
-      process.env.GITHUB_RUN_ATTEMPT = '3';
-      process.env.GITHUB_RUN_ID = '1658821493';
-      process.env.GITHUB_RUN_NUMBER = '3';
-      process.env.GITHUB_SERVER_URL = 'https://github.com';
-      process.env.GITHUB_SHA = 'ffac537e6cbbf934b08745a378932722df287a53';
-      process.env.TEST_RDME_CI = 'true';
-
-      isGHASpy = jest.spyOn(isCI, 'isGHA');
-      isGHASpy.mockReturnValue(true);
-
-      ciNameSpy = jest.spyOn(isCI, 'ciName');
-      ciNameSpy.mockReturnValue('GitHub Actions (test)');
-    });
-
-    afterEach(() => {
-      delete process.env.GITHUB_ACTION;
-      delete process.env.GITHUB_ACTIONS;
-      delete process.env.GITHUB_REPOSITORY;
-      delete process.env.GITHUB_RUN_ATTEMPT;
-      delete process.env.GITHUB_RUN_ID;
-      delete process.env.GITHUB_RUN_NUMBER;
-      delete process.env.GITHUB_SERVER_URL;
-      delete process.env.GITHUB_SHA;
-      delete process.env.TEST_RDME_CI;
-      jest.resetAllMocks();
-    });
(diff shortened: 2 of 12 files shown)


src/lib/fetch.ts
```diff
@@ -84,18 +98,44 @@ function getUserAgent() {
   return `rdme${gh}/${pkg.version}`;
 }
 
+/**
+ * Resolves relative path references for local paths,
+ * otherwise returns the path
+ */
+function normalizeFilePath(opts: FilePathDetails) {
+  if (opts.fileType === 'path') return path.relative('', opts.filePath);
```
kanadgupta, line 106: Why do we need to do this? Say a user passes in a file called `./documentation/rdme.md`. Instead of generating something like this:

```
https://github.com/readmeio/rdme/blob/cb4129d5c7b51ff3b50f933a9c7d0c3d0d33d62c/./documentation/rdme.md
```

we resolve the relative paths and produce this instead:

```
https://github.com/readmeio/rdme/blob/cb4129d5c7b51ff3b50f933a9c7d0c3d0d33d62c/documentation/rdme.md
```

__tests__/helpers/setup-gha-env.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function after() {
+  delete process.env.GITHUB_ACTION;
+  delete process.env.GITHUB_ACTIONS;
+  delete process.env.GITHUB_REPOSITORY;
+  delete process.env.GITHUB_RUN_ATTEMPT;
+  delete process.env.GITHUB_RUN_ID;
+  delete process.env.GITHUB_RUN_NUMBER;
+  delete process.env.GITHUB_SERVER_URL;
+  delete process.env.GITHUB_SHA;
```
erunion, line 37: Is deleting these from within our suite that's running on GH going to cause any problems? Should you reset these to what they were before the overwrites in the `before()` step?

src/cmds/openapi/index.ts
```diff
@@ -235,7 +237,10 @@ export default class OpenAPICommand extends Command {
 
       options.method = 'post';
       spinner.start('Creating your API docs in ReadMe...');
-      return fetch(`${config.get('host')}/api/v1/api-specification`, options).then(res => {
+      return fetch(`${config.get('host')}/api/v1/api-specification`, options, {
```
erunion, line 240: I was very confused why you were passing `filePath` and `fileType` params to our API where they don't exist before realizing that `fetch()` here isn't `fetch` but a wrapper for `fetch`. 🫠

src/lib/fetch.ts
erunion: This `path.relative()` work will have a lot of quirks if `filePath` doesn't look exactly like `./documentation/rdme.md`:

```js
console.log(path.relative('', './documentation/rdme.md'));
console.log(path.relative('', '/documentation/rdme.md'));
console.log(path.relative('', 'documentation/rdme.md'));
console.log(path.relative('', '../documentation/rdme.md'));
```

```
documentation/rdme.md
../../../../../documentation/rdme.md
documentation/rdme.md
../documentation/rdme.md
```

You might want to do something like this:

```js
function sanitizeRelativePath(filePath) {
  return filePath
    .split('/')
    .filter(Boolean)
    .filter(c => c !== '.' && c !== '..')
    .join('/');
}

console.log(path.relative('', sanitizeRelativePath('./documentation/rdme.md')));
console.log(path.relative('', sanitizeRelativePath('/documentation/rdme.md')));
console.log(path.relative('', sanitizeRelativePath('documentation/rdme.md')));
console.log(path.relative('', sanitizeRelativePath('../documentation/rdme.md')));
```

```
documentation/rdme.md
documentation/rdme.md
documentation/rdme.md
documentation/rdme.md
```

src/lib/fetch.ts
```diff
@@ -105,13 +145,35 @@ export default function fetch(url: string, options: RequestInit = { headers: new
...
+  }
+
+  if (isCI()) {
+    headers.set('x-rdme-ci', ciName());
   }
 
   headers.set('x-readme-source', source);
 
+  if (filePath && fileOpts.fileType === 'url') {
+    headers.set('x-readme-source-url', filePath);
```
erunion, line 169: Do we only want to log these in GH like we're doing for `path`?

kanadgupta: I think if someone is running a command in any environment (CI, local script, etc.) and their spec is pointed to a URL, I think it's safe to log that URL as the source URL right? Like if someone were to do something like this:

```sh
rdme openapi https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore.json
```

src/cmds/openapi/index.ts
kanadgupta: 😅 if you have an idea for an alternative pattern, I'm all ears!

__tests__/helpers/setup-gha-env.ts
kanadgupta: Hmm so we've been doing this in our main testbed for a while now and I haven't noticed any issues:
https://github.com/readmeio/rdme/blob/e3c7f22352d6f7e3d24c7fa762df788c905f25e4/__tests__/lib/fetch.test.ts#L29-L35

OH what if that's what's preventing us from writing jest coverage properly 👀 let me play around!

src/cmds/openapi/index.ts
erunion: Aside from renaming the `fetch` wrapper to something else I think it's fine as-is.

src/lib/fetch.ts
kanadgupta: So considering that this normalization only takes place when someone is syncing within a GitHub Actions environment, are they ever not running `rdme` from the root of their repo? This could be famous last words™️ but I can't imagine a scenario in which someone is doing something like this:

```yml
- uses: readmeio/rdme@v8
  with:
    rdme: docs ../documentation/rdme.md
```

Is it even possible to reference a file like that unless you're running some arcane subdirectory workflows? If `rdme` can't find the file, it wouldn't even make the fetch request. But if it can locate the file, we wouldn't want to strip the trailing `../` right? In that case, we might want to do some path resolution from the root of the git repo, but I feel like that's a bridge we should cross if customers ask for it.

erunion: Extreme edge case but what if you run the command with a working directory and then reference a file outside of that?

kanadgupta: Probably but I don't know if people do that in a GitHub Actions runner right?

erunion: What about the `--working-directory` arg on rdme?

I think what you've got is pretty solid fwiw as this work is just for headers I don't think it's the end of the world if we don't catch every weird way people might supply a file or directory.

kanadgupta: Ohhh forgot about that flag. Oof. In that case we might want to do some path resolution from the root of the git repo 😕 I'll think on this more and get back to you!

__tests__/helpers/setup-gha-env.ts
kanadgupta: alas that didn't fix the issues we were seeing in #617 😞 but this approach is a lot cleaner and tests still pass so I'll stick with it: https://github.com/readmeio/rdme/pull/735/commits/eafd73d9be5e4168bedc8f863039d8139af5e51b

src/cmds/openapi/index.ts
kanadgupta: `apiFetch`? `readmeFetch`? `readmeApiFetch`? What do you think?

erunion: `readmeAPIFetch`? And you could consolidate the `config.get('host')` stuff in there too so you don't need to repeat that everywhere.

could also dogfood an `api` SDK too 🥺

kanadgupta: we do some weird fetch stuff like overriding the base URL for proxies, extracting/logging fetch options prior to the request, and setting custom headers that aren't defined in the spec, is that all possible with `api`?

erunion: ah i forgot about those :(

no its not possible with `api` yet

__tests__/lib/fetch.test.ts
```diff
@@ -4,37 +4,14 @@ import { Headers } from 'node-fetch';
...
 import pkg from '../../package.json';
 import fetch, { cleanHeaders, handleRes } from '../../src/lib/fetch';
-import * as isCI from '../../src/lib/isCI';
 import getAPIMock from '../helpers/get-api-mock';
+import { after, before } from '../helpers/setup-gha-env';
 
 describe('#fetch()', () => {
   describe('GitHub Actions environment', () => {
-    let spy: jest.SpyInstance;
+    beforeEach(() => before());
```
RyanGWU82, line 12:
```suggestion
    beforeEach(before);
```

__tests__/lib/fetch.test.ts
```diff
@@ -4,37 +4,14 @@ import { Headers } from 'node-fetch';
...
-      delete process.env.GITHUB_ACTION;
-      delete process.env.GITHUB_ACTIONS;
-      delete process.env.GITHUB_REPOSITORY;
-      delete process.env.GITHUB_RUN_ATTEMPT;
-      delete process.env.GITHUB_RUN_ID;
-      delete process.env.GITHUB_RUN_NUMBER;
-      delete process.env.GITHUB_SHA;
-      spy.mockReset();
-    });
+    afterEach(() => after());
```
RyanGWU82, line 14:
```suggestion
    afterEach(after);
```

RyanGWU82 (approved):
Nothing really to note except for what Jon's already mentioned. The new headers sound potentially pretty useful, thanks for thinking of them.

src/cmds/openapi/index.ts
kanadgupta: I'm gonna rename this in a separate PR to keep this diff confined!

src/lib/fetch.ts
kanadgupta: okay I feel good about this approach: https://github.com/readmeio/rdme/pull/735/commits/3093bcadd5f311bf0e812edf9cc59da124ffb583

let me know what y'all think!

erunion (approved):

PR #737, opened 2023-01-31
fix: page deletion order in `docs:prune`

kanadgupta:
## 🧰 Changes

Noticed two tiny issues in the `docs:prune` command that are fixed in this PR:

- [x] Noticed that folks would see this error if attempting to delete a page that contained children, so I just added a little array sorting into the mix so pages with children are deleted after pages without children.

CleanShot 2023-01-30 at 18 25 36@2x

- [x] Added an extra space between the 🗑️ bin and the confirmation message because yikes:

CleanShot 2023-01-30 at 18 24 42@2x

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/__tests__/cmds/docs/prune.test.ts
+++ b/__tests__/cmds/docs/prune.test.ts
@@ -88,7 +88,7 @@ describe('rdme docs:prune', () => {
         confirm: true,
         version,
       })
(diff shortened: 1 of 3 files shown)


erunion (approved):
seems ok to me

PR #740, opened 2023-01-31
fix(docs): don't merge payloads for PUT requests

### LINKED ISSUE #738: Error when updating doc using `docs` command with `categorySlug` and `parentSlug`
When running the `rdme docs` command with a doc that has the `categorySlug` and/or `parentSlug` attributes in their front matter the following error is returned:

```
Error uploading docs/Test/Child Doc.md:

We couldn't save this doc (received both a `category` and a `categorySlug`, expecting only one).
```

I think this error occurs because the PUT request made in the `updateDoc()` function of the `syncDocsPath` class merges the existing doc payload with the previous docs payload which causes both the category id and the category slug to be included with the payload.

kanadgupta:
| 🚥 Fixes #738 |
| :---------------- |

## 🧰 Changes

For some background, here's the general workflow for when syncing a Markdown file to ReadMe via the `rdme docs` (or `changelogs`/`custompages`) command:

1. Grab the slug from the page
2. Make [a `GET` request](https://docs.readme.com/main/reference/getdoc) to the ReadMe API to confirm that the page exists
3. If the page exists, update the page via [a `PUT` request](https://docs.readme.com/main/reference/updatedoc)
4. If the page does **not** exist, create the page via [a `POST` request](https://docs.readme.com/main/reference/createdoc)

The problem in step 3 is that we take the response body from the `GET` request (called `existingDoc` in the code below) and merge that with the Markdown file's data when making the `PUT` request:

https://github.com/readmeio/rdme/blob/d311af92a0981e3faa4030838d6b4edd8a480dc7/src/lib/syncDocsPath.ts#L107-L111

This isn't really necessary since the API already merges the `PUT` request body with the existing doc, and it's creating adverse side effects when our API expects one field and not another (see #738 for an example).

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/__tests__/cmds/changelogs/index.test.ts
+++ b/__tests__/cmds/changelogs/index.test.ts
@@ -90,7 +90,6 @@ describe('rdme changelogs', () => {
 
       const updateMocks = getAPIMock()
         .put('/api/v1/changelogs/simple-doc', {
-          slug: simpleDoc.slug,
           body: simpleDoc.doc.content,
           lastUpdatedHash: simpleDoc.hash,
           ...simpleDoc.doc.data,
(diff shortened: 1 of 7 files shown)


erunion (approved):

garrett-wade (approved):
Looks good, had the exact same code in a branch ready to go :)

PR #746, opened 2023-02-10
Semantic release

kanadgupta:
| 🚥 Fixes ISSUE_ID |
| :---------------- |

## 🧰 Changes

Describe in detail what this PR is for.

## 🧬 QA & Testing


CHANGELOG.md
```diff
@@ -1,3 +1,66 @@
+# [8.6.0-next.7](https://github.com/readmeio/rdme/compare/v8.6.0-next.6...v8.6.0-next.7) (2023-02-13)
```
kanadgupta, line 8: there are a few formatting discrepancies in this file now but I really care to be honest

.releaserc.yml
```diff
@@ -0,0 +1,30 @@
...
+  # the package*.json files before we commit the changes
+  - '@semantic-release/npm'
+  - [
+      '@semantic-release/git',
+      {
+        # TODO: once docker builds are a thing, we're going to have to manually add the assets
+        # since we're also updating the action.yml file with every release!
+        # Uncomment this line below once those changes are live:
+        # assets: ['action.yml', 'CHANGELOG.md', 'package.json', 'package-lock.json'],
+        message: "build(release): 🚀 v${nextRelease.version} 🦉\n\n${nextRelease.notes}\n[skip ci]",
```
kanadgupta, line 19: `[skip ci]` is a special phrase, see https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

erunion (approved):
nice

RyanGWU82:
A couple of general questions:

1. What's the value of having a prerelease process on the `next` branch?
2. Have we confirmed that the code pushes from `semantic-release` don't themselves trigger new runs of `semantic-release`? (Let's make sure this won't cause an infinite loop 😭)
3. Have you confirmed this actually works as a GitHub Action? We weren't able to get that working on `readme` because GitHub doesn't allow code pushes from inside GitHub Actions, even if the `GITHUB_TOKEN` would otherwise have access privileges to do this. If it does work, do you know how/why it works since it didn't on `readme`? (Could we apply the same "fix" to `readme`? 🤔)

RyanGWU82 (approved):
Great documentation! 🙇🏼

kanadgupta:
> 1. What's the value of having a prerelease process on the `next` branch?

A couple of reasons!

1. We can actually do a global install (i.e., `npm i -g rdme`) locally and play around with prerelease version in a variety of different directories to confirm that changes work as expected (this is something I've wanted for a while)
2. Since we're going to publish every Docker build to the GitHub container registry (inspired by [this repo](https://github.com/jacobtomlinson/gha-find-replace/blob/4d085e89079e4d2e6971679ebc87da71492aa462/action.yml#L27)), we need an end-to-end technique to test out the GitHub Action before tagging official releases... in other words, we need to use a tagged Docker container release to properly test GitHub Actions 😬 not ideal but the pros should hopefully outweigh the cons 🤞🏽 

> 3. Have you confirmed this actually works as a GitHub Action? We weren't able to get that working on `readme` because GitHub doesn't allow code pushes from inside GitHub Actions, even if the `GITHUB_TOKEN` would otherwise have access privileges to do this. If it does work, do you know how/why it works since it didn't on `readme`? (Could we apply the same "fix" to `readme`? 🤔)

Yeah I remember those issues... will report back once this workflow actually runs in `next` and `main`, but I've tried basically everything I can think of on this branch (protecting this branch, actually tagging releases, etc.) and the `GITHUB_TOKEN` appears to be working fine? Maybe GitHub has since made some fixes to their Actions permissioning so we can finally do this now 🧐

@RyanGWU82

RyanGWU82:
Thanks @kanadgupta. The issue with GHA permissions was that you couldn't use GHA to actually make commits on a git branch. You could interact with the API as much as you wanted (for tagging, replying to issues, etc) but you couldn't actually save the commit to update the version number or the changelog. Were you able to test it on a branch and have it actually make a commit on that branch?

It's been over a year since I looked at that, so maybe they got it working in a newer version of semantic-release?

kanadgupta:
@RyanGWU82 yeah there are a few examples on this branch if you want to look, like this one: https://github.com/readmeio/rdme/commit/05af28561eab4c594795def6ca8cb3e094ac0ccd

I also tried adding branch protections and it still appears to work 🧐

PR #758, opened 2023-02-14
chore: npm/git ignore `.env`

kanadgupta:


RyanGWU82 (approved):

.npmignore
```diff
@@ -1,4 +1,5 @@
 __tests__/
```
erunion, line 1: @kanadgupta @RyanGWU82 It might be safer to change `.npmignore` to be opt-in. I think something like this would do it:

```
 *
!bin/
!config/
!dist/
action.yml
!CHANGELOG.md
!CONTRIBUTING.md
!LICENSE
!MAINTAINERS.md
!package-lock.json
!package.json
!README.md
```

I believe you can do `npm publish --dry-run` to verify this though.

kanadgupta: I'm a bit nervous nervous about this approach — what if we add a new file that's required and forget to update this file? Is there any way of us knowing that it won't work until we actually publish -> local install?

erunion: there's not, unfortunately

PR #759, opened 2023-02-14
feat: docker

kanadgupta:
| 🚥 Fixes RM-6538 |
| :---------------- |

## 🧰 Changes

This is a tiny PR to get our GitHub Action using Docker instead of a composite action. It takes **substantially** longer to install/build than the composite action approach in its current form because it's rebuilding the container with every workflow run, but I'll get the container registry publishing going in a separate PR (once this and #746 are in `next`) so we can really speed things up 🚀

## 🧬 QA & Testing

Do our [GitHub Actions dry runs](https://github.com/readmeio/rdme/actions/runs/4179287016/jobs/7239056307) continue to work as expected?


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM node:16
+
+COPY . /rdme-docker
```
kanadgupta, line 3: Not sure what the best practices are regarding directory stuff like this... feedback appreciated 😭

RyanGWU82: It varies a lot AFAIK. I've seen `/app` before sometimes, that may be what we use in our main app. You could use `/rdme` if you want.

RyanGWU82 (approved):

domharrington (commented):
> It takes substantially longer to install/build than the composite action approach in its current form because it's rebuilding the container with every workflow run

So will this actually be faster when we start using the registry? We should probably get that working e2e before committing to this approach just to make sure the benefits actually outweigh it here? It's not like this is a complicated project, and npm install can't take _that_ long. Where as `node:16` which is the base image we're using here is a not-exactly-lean 340MB:

```sh
$ docker pull node:16

$ docker images --filter=reference="node:16"
REPOSITORY   TAG       IMAGE ID       CREATED         SIZE
node         16        ce3d204ff857   2 minutes ago   340MB
```

I dont know either way which approach is better, but would be interested to prove it out more to ensure benefits. We could also investigate using one of the leaner docker images like 16-alpine, which is a slightly more palatable 42.3MB:

```sh
$ docker images --filter=reference="node:16-alpine"
REPOSITORY   TAG         IMAGE ID       CREATED          SIZE
node         16-alpine   1621ddffc775   39 seconds ago   42.3MB
```

kanadgupta:
> We should probably get that working e2e before committing to this approach just to make sure the benefits actually outweigh it here?

Yep, that's the point of the prerelease pipeline in #746 🙃 and didn't know about `node:16-alpine`, will look into that, thanks!

domharrington:
Sorry I didnt know about that! Hoping it all speeds up nicely 🚀

PR #760, opened 2023-02-14
chore: update JSON schema for github workflow files

kanadgupta:
## 🧰 Changes

In our test suite, we have a copy of the JSON schema for GitHub Actions workflows, which is used to validate the GitHub Actions workflow files that we generate as part of [our GitHub Actions onboarding](https://docs.readme.com/main/docs/rdme#quick-start).

The source can be found here: https://json.schemastore.org/github-workflow.json

You can view the full file history here: https://github.com/SchemaStore/schemastore/commits/master/src/schemas/json/github-workflow.json

## 🧬 QA & Testing

If tests pass we should be good to go!


erunion (approved):

erunion:
do you think it's worth adding a script to download this? or maybe writing a test that pulls the latest down and fails if our copy doesnt match?

kanadgupta:
@erunion went down a bit of a rabbit hole and added a `pretest` check for this, along with a script to refresh it! Let me know what you think 👀

src/.sink.d.ts
```diff
@@ -1,3 +1,4 @@
 // These packages don't have any TS types so we need to declare a module in order to use them.
+declare module '@readme/eslint-config';
```
erunion, line 2: Since this doesn't need to get loaded for the whole app you might be able to move this line into the top of the new script you made. That might work?

erunion (approved):

src/.sink.d.ts
kanadgupta: weird, I thought I needed this but I think I'm able to remove this line entirely and not re-add it anywhere. let me try that

erunion: Might also want to exclude that script from being added into `dist/` if `npm run build` is adding it there now.

kanadgupta: looks like we already exclude the `bin` directory:

https://github.com/readmeio/rdme/blob/d2b96d4474003e04b8a7543bfb6de075ef02aa54/tsconfig.json#L17

kanadgupta: cool looks like tests/build are passing without this line! weird

PR #764, opened 2023-02-15
perf(docker): build executable

kanadgupta:
## 🧰 Changes

I initially opened this PR with the goal of automatically publishing these docker images to the GitHub Container Registry, but in testing it I discovered that these images were so large that the GitHub Action time savings were basically a wash (the benefits were only realized in workflows that ran `rdme` more than once due to the image caching).

I decided to try and use this PR to instead optimize our images as much as possible, so now we create an executable using [`pkg`](https://github.com/vercel/pkg) and create a tiny image that only runs that. Was able to reduce our image size by ~90% 🥳 the build time is higher now, but once we publish to the registry our users shouldn't have to deal with that.

**Old image size**: 639.63267 MB

**New image size**[^1]: 71.004255 MB 📉 

## 🧬 QA & Testing

If the GitHub Action dry runs continue to work, we should be good to go!

[^1]: This is the size of the image when I built it on my Mac but I imagine the GitHub Action runner image is roughly the same?


erunion (approved):
nice

PR #777, opened 2023-03-04
feat: container registry

kanadgupta:
| 🚥 Fixes RM-6538 |
| :---------------- |

## 🧰 Changes

The moment we've been waiting for y'all: enhancing our `semantic-release` workflow to automatically publish docker images to the GitHub Container Registry!

There are a bunch of moving parts, but if you look at the Linear ticket, I've added comments with links to a couple internal videos that I created

## 🧬 QA & Testing

I did a couple of test pushes to the Container Registry and confirmed that a working, containerized version of `rdme` is pushed to the registry!

- You can see the release workflow here: https://github.com/readmeio/rdme/actions/runs/4350052453/jobs/7600379201
- And the corresponding image in the container registry here: https://github.com/readmeio/rdme/pkgs/container/rdme/75372142

And the performance gains are pretty incredible. Here's [the before](https://github.com/readmeio/rdme/actions/runs/4328524620/jobs/7558303832):

image

And here's [the after](https://github.com/readmeio/rdme/actions/runs/4328516747/jobs/7596996985):

image

There are a few aspects of this process that we can only test once we merge into `next`, but that's why we're doing pre-releases now amirite?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,13 @@ jobs:
           path: oas-examples-repo
           repository: readmeio/oas-examples
 
+      - name: Update action.yml to build locally
+        uses: jacobtomlinson/gha-find-replace@v3
+        with:
+          find: 'image:.*'
+          replace: "image: 'Dockerfile'"
+          include: rdme-repo/action.yml
+
       - name: Run `openapi:validate` command
         uses: ./rdme-repo/
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,12 +52,12 @@ jobs:
           repository: readmeio/oas-examples
 
       - name: Run `openapi:validate` command
-        uses: readmeio/rdme@v8
+        uses: ./rdme-repo/
         with:
           rdme: openapi:validate oas-examples-repo/3.1/json/petstore.json
 
       # Docs: https://rdme-test.readme.io
       - name: Run `openapi` command
-        uses: readmeio/rdme@v8
+        uses: ./rdme-repo/
         with:
           rdme: openapi oas-examples-repo/3.1/json/petstore.json --key=${{ secrets.RDME_TEST_PROJECT_API_KEY }} --id=${{ secrets.RDME_TEST_PROJECT_API_SETTING }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,12 @@ jobs:
           path: oas-examples-repo
           repository: readmeio/oas-examples
 
-      - name: Update action.yml to build locally
+      # For every GitHub Action release, we deploy our Docker images
+      # to the GitHub Container Registry for performance reasons.
+      # Instead of testing against the pre-built image, we can build
+      # an image from the currently checked-out repo contents by doing a
+      # li'l update in the action.yml file to point to the current Dockerfile.
+      - name: Replace Docker image value in action.yml
         uses: jacobtomlinson/gha-find-replace@v3
         with:
           find: 'image:.*'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,12 +52,12 @@ jobs:
           repository: readmeio/oas-examples
 
       - name: Run `openapi:validate` command
-        uses: ./rdme-repo/
+        uses: readmeio/rdme@496c511ce3a46b03e5726858271dd22d4cdc12d1
         with:
           rdme: openapi:validate oas-examples-repo/3.1/json/petstore.json
 
       # Docs: https://rdme-test.readme.io
       - name: Run `openapi` command
-        uses: ./rdme-repo/
+        uses: readmeio/rdme@496c511ce3a46b03e5726858271dd22d4cdc12d1
         with:
           rdme: openapi oas-examples-repo/3.1/json/petstore.json --key=${{ secrets.RDME_TEST_PROJECT_API_KEY }} --id=${{ secrets.RDME_TEST_PROJECT_API_SETTING }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,12 +52,12 @@ jobs:
           repository: readmeio/oas-examples
 
       - name: Run `openapi:validate` command
-        uses: readmeio/rdme@496c511ce3a46b03e5726858271dd22d4cdc12d1
+        uses: readmeio/rdme@v8
         with:
           rdme: openapi:validate oas-examples-repo/3.1/json/petstore.json
 
       # Docs: https://rdme-test.readme.io
       - name: Run `openapi` command
-        uses: readmeio/rdme@496c511ce3a46b03e5726858271dd22d4cdc12d1
+        uses: readmeio/rdme@v8
         with:
           rdme: openapi oas-examples-repo/3.1/json/petstore.json --key=${{ secrets.RDME_TEST_PROJECT_API_KEY }} --id=${{ secrets.RDME_TEST_PROJECT_API_SETTING }}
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -13,14 +17,48 @@ jobs:
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          # For every release we add 2-3 tags to the docker build:
+          # 1. A semver tag (e.g., 8.0.0, 8.0.0-next.0)
+          # 2. A branch tag (e.g., next, main)
+          # 3. A `latest` tag (only on the main branch)
+          tags: |
+            type=semver,pattern={{version}},value=${{ steps.release.outputs.nextVersion }}
+            type=ref,event=branch
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
```
erunion, line 60: What's this doing? Setting `latest` to `main` or to the commit sha?

bin/docker.js
erunion, line 1: You're calling `pkg` [here](https://github.com/readmeio/rdme/blob/6c9f8ba07801c2eddf25ef8d25bb90568c1029ae/Dockerfile#L5) in the Dockerfile but do you want to pin that in `package.json` as a small safety/security precaution? If something ever happens to the latest release of that package and we build an image at the same time I wouldn't want us to unexpectedly break.

.github/workflows/ci.yml
```diff
@@ -51,6 +51,18 @@ jobs:
...
+      # to the GitHub Container Registry for performance reasons.
+      # Instead of testing against the pre-built image, we can build
+      # an image from the currently checked-out repo contents by doing a
+      # li'l update in the action.yml file to point to the current Dockerfile.
+      - name: Replace Docker image value in action.yml
+        uses: jacobtomlinson/gha-find-replace@v3
+        with:
+          find: 'image:.*'
+          replace: "image: 'Dockerfile'"
+          include: rdme-repo/action.yml
```
erunion, line 64: Why not call `bin/set-action-image.sh` instead?

bin/verify-clis.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Verify existence of gh and docker CLIs
+# https://stackoverflow.com/a/677212
+set -e
+if ! command -v docker &> /dev/null
+then
+    echo "docker CLI could not be found"
+    exit 1
+fi
+if ! command -v gh &> /dev/null
```
erunion, line 11: What are we using `gh` for?

bin/docker.js
```diff
@@ -0,0 +1,95 @@
...
+    return process.exit(1);
+  }
+}
+
+/**
+ * Runs command and logs all output
+ */
+async function runCmd(cmd) {
+  // https://stackoverflow.com/a/63027900
+  const execCmd = exec(cmd);
```
RyanGWU82, line 32: `child_process.exec()` takes a string that it passes to the shell. However that's not particularly safe, and we're not doing anything to sanitize the input when building the shell commands in `run()`.

> Never pass unsanitized user input to this function. Any input containing shell metacharacters may be used to trigger arbitrary command execution.

Source: https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

Can we switch these commands to `child_process.execFile` and pass an array of arguments? https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback

bin/docker.js
```diff
@@ -0,0 +1,95 @@
...
+const util = require('util'); // eslint-disable-line unicorn/import-style
+const exec = util.promisify(require('child_process').exec);
+
+/**
+ * Retrieves and parses the docker image metadata
+ * @see {@link https://github.com/docker/metadata-action}
+ */
+function getMetadata() {
+  try {
+    const raw = process.env.DOCKER_METADATA_OUTPUT_JSON;
```
RyanGWU82, line 12: Do you have an example of what this looks like? I'm not sure I understand what labels and tags we might be applying here. (particularly the labels—I do see the tags in `release.yml`)

RyanGWU82 (changes requested):
I don't feel like I have a good grasp of the big picture of everything that happens during a release here, but if you've tested it and you're comfortable with it, then that's a good sign. Is there documentation for the build/release process and what's happening under the hood?

Can't approve it because of the unsanitized input to shell commands, but once that's fixed I'm 👍🏼.

.github/workflows/ci.yml
kanadgupta: ah I hadn't thought of that — this is pretty similar. issue is we'd need to add a step to install deps since we use `js-yaml` and that doesn't feel totally worth it 😕

.github/workflows/release.yml
kanadgupta: say we're on the `main` branch and this workflow is tagging the `8.6.0` release, it'll add the following three tags to the current docker build:
1. `8.6.0`
2. `main` (it'll be `next` if on the `next` branch)
3. `latest` (this will only be added on the `main` branch, which is what that line is doing)

i talk about this here but let me know if I should fluff this out at all:
https://github.com/readmeio/rdme/blob/3b153c84159182499e7dd66fcc276cfcb05e1bfd/.github/workflows/release.yml#L53-L56

bin/verify-clis.sh
kanadgupta: we use it to create the draft GitHub release:
https://github.com/readmeio/rdme/blob/3b153c84159182499e7dd66fcc276cfcb05e1bfd/.releaserc.yml#L42

bin/docker.js
kanadgupta: So I might be misunderstanding but I'm not sure arbitrary command execution is much of an issue here — I already do some sanitization checks here to ensure that the env variable input is valid (and this is really the only "user" input): https://github.com/readmeio/rdme/blob/52d80d4640f2900c474c345dda4382cdee0eb40c/bin/docker.js#L16-L21

And beyond that, everything is mostly hardcoded strings to construct the docker CLI commands and call the `runDockerCmd` (f.k.a. `runCmd`) function (and I've ran through the commands locally a bunch to confirm that they're valid).

In any case, I refactored this in https://github.com/readmeio/rdme/pull/777/commits/1b07bb7def65224317d2dc1f38e88516a46af38f to use `execFile`, TIL! It's **much** more readable now too so thanks for the suggestion 🎊

bin/docker.js
kanadgupta: Yep! Here's an example:

```json
{
  "tags": ["ghcr.io/readmeio/rdme:8.6.0-next.16", "ghcr.io/readmeio/rdme:next"],
  "labels": {
    "org.opencontainers.image.title": "rdme",
    "org.opencontainers.image.description": "ReadMe's official command-line interface (CLI) and GitHub Action 🌊",
    "org.opencontainers.image.url": "https://github.com/readmeio/rdme",
    "org.opencontainers.image.source": "https://github.com/readmeio/rdme",
    "org.opencontainers.image.version": "8.6.0-next.16",
    "org.opencontainers.image.created": "2023-03-21T21:10:25.208Z",
    "org.opencontainers.image.revision": "064e935860133f2fbe7ba0d6cb84507c65397689",
    "org.opencontainers.image.licenses": "MIT"
  }
}
```

I've created a gist that contains this and linked to it in the comments: https://github.com/readmeio/rdme/pull/777/commits/6b5c0c46498572375b56df9cdce5232693ccae69

erunion (approved):
:nice-rgb:

bin/docker.js
RyanGWU82: Looks good! My thinking was the tags, which come from git branch names and those could be anything. It looks like git does have some sanitization (I tried a semicolon and got `fatal: 'foo; reboot' is not a valid branch name`) but I could imagine there are ways around this if someone was motivated to try hard enough.

RyanGWU82 (approved):

PR #788, opened 2023-03-31
fix(docker): safer argument parsing for GitHub Actions

### LINKED ISSUE #787: Filename in quotes no longer working
Hi,

Our team was using the rdme openapi tool to automatically sync our docs on pull requests. This broke for us recently because filenames in quotes are no longer supported.

Previously this worked:
```- uses: readmeio/rdme@v8
        with:
          rdme: openapi "<file_name>" --key ${{ secrets.README_DEVELOPMENT_API_KEY }} --id ${{ secrets.README_DEV_ID }}
```

We fixed it on our side by removing the quotes but in the future please be more careful about making breaking changes, thanks!

kanadgupta:
| 🚥 Fixes #787 |
| :---------------- |

#### The Problem 😵‍💫

Someone wrote in in #787 about an issue where their workflows started breaking due to them wrapping their file names in quotation marks. Normally our args parser ([`command-line-args`](https://github.com/75lb/command-line-args)) would handle quotation marks just fine, but because of our new Docker setup and how we have the user pass in all their `rdme` arguments into a single GitHub Actions input, quotation marks are escaped by the time `rdme` receives it 😬

To put it a different way, this was the equivalent of running something like this:

CleanShot 2023-04-04 at 17 34 02@2x

#### The Solution 💭 

This was a pain to debug and fix (thanks @RyanGWU82 and @domharrington for pairing on this 😮‍💨), but I got something going that I'm happy with! Now, when we invoke the `rdme` Docker image via the GitHub Action runner, it's basically running this:

```sh
rdme docker-gha "openapi:validate \"https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore-simple.json\""
```

The `docker-gha` argument isn't documented nor exposed to users, but is basically an internal escape hatch so we can properly consume `rdme` arguments that are sent to us via a single string.

As part of this, I did the following:
- [x] Refactors our `Dockerfile` and our `action.yml` files so we're now using `rdme` as the executable and passing in the `rdme` command via good ol' fashioned `docker run` arguments.

> To put this a different way, here's the before and after if we were to build the `rdme` Docker image locally and pass arguments into it:
> 
> **Before** (using the `INPUT_RDME` environmental variable):
> 
> ```sh
> docker run -i -t --env INPUT_RDME='openapi:validate https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore-simple.json' ghcr.io/readmeio/rdme:dev
> ```
> 
> **After** (using standard `docker run` arguments):
> 
> ```sh
> docker run -i -t --env DEBUG='rdme*' ghcr.io/readmeio/rdme:dev openapi:validate https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore-simple.json
> ```

- [x] Added logic our main `rdme` index.ts file to check for this new `docker-gha` command and if it exists, parse out the arguments using [`string-argv`](https://github.com/mccormicka/string-argv).
- [x] Added corresponding test coverage, both in Jest and in our GitHub Actions dry runs.

## 🧬 QA & Testing

If tests pass we should be good to go! If you can think of any edge cases with quotation marks and string serialization I might be missing, please let me know 🥺

To get a sense of what is going on here, feel free to run this locally:

```sh
# Build + install deps
npm ci && npm run build

# simulate the command that is being invoked by the GitHub Actions runner
bin/rdme docker-gha "openapi:validate \"https://github.com/readmeio/oas-examples/blob/main/3.0/json/petstore-simple.json\""
```


action.yml
```diff
@@ -14,3 +14,6 @@ outputs:
 runs:
   using: docker
   image: docker://ghcr.io/readmeio/rdme:8.6.0
+  args:
+    - docker-gha
```
kanadgupta, line 18: This is a new internal argument we send to indicate to the `rdme` Docker image that the command is going to be a single string as opposed to an array of strings.

src/index.ts
```diff
@@ -45,6 +45,36 @@ export default function rdme(processArgv: NodeJS.Process['argv']) {
...
+   *
+   * The GitHub Actions runner will send this to the `rdme` Docker image:
+   * ['openapi "petstore.json"']
+   *
+   * To distinguish these, we have a hidden `docker-gha` argument which we check for to indicate
+   * when arguments are coming from the GitHub Actions runner.
+   * This logic checks for that `docker-gha` argument and parses the second string
+   * into the arguments array that `command-line-args` is expecting.
+   */
+  if (rawProcessArgv.length === 2 && rawProcessArgv[0] === 'docker-gha') {
```
kanadgupta, line 67: We now check for that `docker-gha` argument before parsing out the second string and converting it to something that `command-line-args` can process.

Dockerfile
```diff
@@ -8,4 +8,4 @@ FROM alpine:3.14
 
 COPY --from=builder /rdme/exe /exe
 
-ENTRYPOINT ["sh", "-c", "/exe/rdme $INPUT_RDME"]
+ENTRYPOINT ["/exe/rdme"]
```
kanadgupta, line 11: Not only this this a lot cleaner but we can now use good ol' `docker run` arguments, but GitHub recommends[^1] accessing inputs via arguments and not environmental variables [in their own docs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-specifying-inputs):

CleanShot 2023-04-04 at 18 07 29@2x

[^1]: They technically imply that it's impossible to access the `INPUT_` environmental variables from the Docker container action, but that's definitely wrong — `rdme@8.6.0`  wouldn't be working at all in that case lol. The takeaway here is that we should be passing inputs via arguments and not environmental variables.

src/index.ts
```diff
@@ -45,6 +45,36 @@ export default function rdme(processArgv: NodeJS.Process['argv']) {
...
+    processArgv = rawProcessArgv[1].split(' ').map(arg => {
```
erunion, line 68: Splitting on spaces here might be problematic.

```
openapi:validate "__tests__/__fixtures__/petstore simple weird version.json"
```

kanadgupta: ugh shit yeah. thoughts on using this? https://www.npmjs.com/package/string-argv

erunion: Looking at the regex it has for handling these cases, yeah we should use this.

https://github.com/mccormicka/string-argv/blob/master/index.ts#L14

kanadgupta: Cool, will refactor tomorrow!

kanadgupta: added! god this library rules. zero deps, fully typed, does exactly what it's supposed to and hasn't required an update in 4+ years :chefs-kiss:

erunion (approved):
hell of a bug + fix

action.yml
domharrington: This is a pretty neat/sufficient solution. Annoying that it's necessary, but it's not too bad.

src/index.ts
domharrington: Hopefully this is something we can remove down the line! But nice fix for now.

domharrington (approved):
Good job!

.github/workflows/ci.yml
```diff
@@ -68,8 +68,18 @@ jobs:
         with:
           rdme: openapi:validate oas-examples-repo/3.1/json/petstore.json
 
+      - name: Run `openapi:validate` with filename in quotes
+        uses: ./rdme-repo/
+        with:
+          rdme: openapi:validate "oas-examples-repo/3.1/json/petstore.json"
```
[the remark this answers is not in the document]
kanadgupta: Correct, both tests added to this file check for the breakage in the most recent release!

rafegoldberg (approved):
Frustrating bug; interesting fix. Changes look good to me.

RyanGWU82 (approved):
Thanks Kanad!

PR #820, opened 2023-06-06
chore(eslint): consistenly sorting on interface props

erunion:
## 🧰 Changes

Our core ESLint standards now includes the [eslint-plugin-typescript-sort-keys](https://www.npmjs.com/package/eslint-plugin-typescript-sort-keys) plugin for enforcing consistent sorting of TS `interface` props.


kanadgupta (approved):
ty

PR #839, opened 2023-07-31
docs: use updated callout syntax

kanadgupta:
No material changes here, just a small syntax update to reflect the updated GFM callout format: https://github.com/orgs/community/discussions/16925

Per the author:

> A new syntax, `[!NOTE]`, has been added, which will gradually replace the old one. However, the old syntax will continue to work for some time.


kanadgupta:
@erunion FYI!

PR #851, opened 2023-08-02
docs(openapi): add postman warning

t-tullis:
| 🚥 Resolves CX-364 |
| :------------------- |

## 🧰 Changes

Describe in detail what this PR is for.

## 🧬 QA & Testing


src/cmds/openapi/validate.ts
kanadgupta, line 1: Mind reverting this change?

src/lib/prepareOas.ts
```diff
@@ -211,6 +211,34 @@ export default async function prepareOas(
...
+          'You are inspecting a Postman collection. This feature is currently experimental. For more information, visit our docs here: https://docs.readme.com/main/docs/openapi#the-api-reference',
+        );
+        break;
+      case 'openapi:convert':
+        warn(
+          'You are attempting to convert a Postman collection to an OpenAPI file. This feature is currently experimental. For more information, visit our docs here: https://docs.readme.com/main/docs/openapi#the-api-reference',
+        );
+        break;
+      default:
+        debug(`Type ${chalk.yellow('rdme help')} to see all commands`);
```
kanadgupta, line 238: don't think you need this `debug` statement, mind removing it?

src/lib/prepareOas.ts
```diff
@@ -211,6 +211,34 @@ export default async function prepareOas(
     debug('spec bundled');
   }
 
+  // Checks to see if the selected file is a postman collection and throws a warning in the console depending on the command.
+  if (definitionVersion.specification === 'postman') {
+    switch (command) {
+      case 'openapi':
+        warn(
+          'You are attempting to upload a Postman collection. This feature is currently experimental. For more information, visit our docs here: https://docs.readme.com/main/docs/openapi#the-api-reference',
```
kanadgupta, line 219: small refactoring suggestion: mind creating a separate variable above (called `postmanWarning` or something like that) and re-using that in all of these `warn` statements?

src/lib/prepareOas.ts
```diff
@@ -211,6 +211,34 @@ export default async function prepareOas(
     debug('spec bundled');
   }
 
+  // Checks to see if the selected file is a postman collection and throws a warning in the console depending on the command.
+  if (definitionVersion.specification === 'postman') {
+    switch (command) {
+      case 'openapi':
```
kanadgupta, line 217: great job here with this `if`/`switch` logic — super readable and it's in the perfect spot 👏🏽

kanadgupta (changes requested):
great work so far, for real! the code is in the exact spot it should be and you've absolutely nailed the logic — i have a few small suggestions below, and i think it might be good to add a test to ensure that it shows up when we want it to. happy to pair on this!

PR #854, opened 2023-08-09
chore: clean up tsconfig files

kanadgupta:
## 🧰 Changes

Been doing a few deep-dives to understand tsconfigs and saw a few things that we could probably remove? I also stumbled upon [the official recommended tsconfig for Node 14](https://github.com/tsconfig/bases/blob/main/bases/node14.json) and brought in some properties from there.

The biggest change is bumping our `target` to `ES2020`. Not sure if this means anything but it yields a smaller dist (~32% smaller), yay!

See below for the before and after:

CleanShot 2023-08-09 at 16 58 27@2x
CleanShot 2023-08-09 at 16 57 58@2x

## 🧬 QA & Testing

Is this okay to do? Tests seem to pass so I think it's fine?


tsconfig.json
```diff
@@ -5,14 +5,15 @@
     "downlevelIteration": true,
     "esModuleInterop": true,
     "experimentalDecorators": true,
-    "lib": ["es2020", "ES2021.Intl"],
-    "noImplicitAny": true,
+    "forceConsistentCasingInFileNames": true,
+    "lib": ["ES2020", "ES2021.Intl"],
+    "module": "CommonJS",
```
kanadgupta, line 10: I tried setting this to `node16` since [that's what's recommended](https://github.com/tsconfig/bases/blob/main/bases/node14.json) but it broke the resolution for a handful of dependencies that ship with both CJS/ESM builds. I love this ecosystem 😭

erunion: you might want to build out some integration tests that run the compiled bin, maybe with `--help`?, to make sure that it works on all versions of node

kanadgupta: does this count? https://github.com/readmeio/rdme/blob/198d4c42c8eca7a723e2ba8e7842cf87f8f08485/__tests__/bin.test.ts#L8

erunion: oh sick, i forgot we had those.

erunion (approved):

PR #855, opened 2023-08-09
fix(openapi/inspect): small url formatting error

kanadgupta:
## 🧰 Changes

Noticed a tiny formatting error when running `openapi:inspect` against an OpenAPI 3.0 definition:

CleanShot 2023-08-09 at 16 17 40@2x

I made the type a little stricter and tweaked the logic a bit so now it looks like this:

CleanShot 2023-08-09 at 16 22 31@2x

I don't love how the URL and this disclaimer text have the same formatting, but this is probably good enough for now.

## 🧬 QA & Testing

Do tests pass and does this change look good?


erunion (approved):
ope

t-tullis (approved):
lgtm!

Issue #801, opened 2023-04-18
v9 roadmap


kanadgupta:
## `v9`

- [x] some sort of deprecation notice for `rdme openapi`? [#1106](https://github.com/readmeio/rdme/pull/1106)
- [x] https://github.com/readmeio/rdme/pull/1099
- [x] https://github.com/readmeio/rdme/pull/1098
- [x] https://github.com/readmeio/rdme/pull/1100
- [x] remove old command reference language from main `README`, in favor of new autogenerated CLI reference docs in `documentation/commands.md`
- [x] get `dist-gha/` directory to update with every new release https://github.com/readmeio/rdme/commit/f42392b5141d82df2acfa10df82989fe4832b36f
- [x] https://github.com/readmeio/rdme/pull/1072
- [x] #962, #1068
  - [x] update GHA build process to reflect `oclif` setup
- [x] update `MAINTAINERS.md`
- [x] Auto-publish GitHub releases (e.g., remove `gh` CLI usage) 331d28bc1e9f0f2b57a4d8d39f2ebaf6bf64896a
- [x] Make sure JSON import assertions are future-proof (or aren't used at all) https://github.com/readmeio/rdme/pull/993
- [x] #889
- [x] Migrate to ESM (#856)
  - [x] Do some QA to ensure no Node `ExperimentalWarning`s show up https://github.com/readmeio/rdme/pull/901
- [x] Drop support for Node <20 ([release schedule](https://endoflife.date/nodejs)) https://github.com/readmeio/rdme/pull/900 https://github.com/readmeio/rdme/pull/1085
  - [x] Bump lockfile to v3
- [x] Remove `docs:edit` command https://github.com/readmeio/rdme/pull/902
- [x] Remove `oas` command https://github.com/readmeio/rdme/pull/902
- [x] Remove `swagger` command https://github.com/readmeio/rdme/pull/902
- [x] _(Probably separate from the ESM migration)_ look into any clean-up/modernizations we can do in the `tsconfig` https://github.com/readmeio/rdme/pull/905
- [x] flip the `isPublic` flag in the `versions:create` and `versions:update` commands to be `hidden` so it's consistent with the other flags and with our [API endpoints](https://docs.readme.com/main/reference/createversion) https://github.com/readmeio/rdme/pull/906
- [x] https://github.com/readmeio/rdme/pull/895
- [x] #857
[...]

PR #856, opened 2023-08-09
feat: convert to ESM (breaking change)

kanadgupta:
| 🚥 Partially addresses #801 |
| :------------------- |

## 🧰 Changes

This PR refactors the entire repository over to full ESM. This entailed a few different refactors:

- [x] Using `.js` file extensions for relative imports (really wish I had done this _**after**_ discovering [`eslint-plugin-require-extensions`](https://github.com/solana-labs/eslint-plugin-require-extensions)) 😭
- [x] Updated a bunch of dependencies to their latest ESM counterparts
  - (with the exception of `ora`, see the outstanding issues below)
  - Some of our upstream dependencies now use native Node 18 `fetch` so a test or two had to be updated accordingly. Eventually we'll want to migrate away from `nock` entirely.
- [x] Setting `module: NodeNext` in our `tsconfig.json` and `type: module` in our `package.json`
- [x] In order to build the single-file executable binaries used in our Docker/GitHub Actions builds, we now compile this entire repository into CJS using Rollup before sending it into `pkg` 😵‍💫 (see https://github.com/vercel/pkg/issues/1291)
  - This obviously isn't ideal, but `pkg` handles a lot of the madness of compiling into single-file executables better than anything else I've tried ([see here for more context](https://github.com/readmeio/rdme/pull/856#issuecomment-1673332625))
  - A huge benefit here is that our Docker image sizes should be around 25% smaller 📉

<details>

<summary><b>Outstanding tasks/issues</b></summary>

Below is a running list of outstanding work — the `ora` issue is mostly out of our hands, but I'll look into the `ExperimentalWarning` in a follow-up PR.

- [ ] Unable to update to the latest version of `ora` due to segmentation faults at runtime, see https://github.com/sindresorhus/ora/issues/229 for more information
- [ ] Figure out if there's a non-shitty way to hide this warning (**edit**: I tried fixing this in https://github.com/readmeio/rdme/pull/856/commits/eeeabd950d56bf93c7430ce903d395541642f575 but it proved to be problematic for a variety of reasons):

```
(node:92816) ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:92816) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
```
- [x] Unable to properly load in some of our packages
  - [x] `oas` (see https://github.com/readmeio/oas/pull/796)
  - [x] `oas-normalize` (see https://github.com/readmeio/oas-normalize/pull/294)
- [x] `tsconfig` cleanup (e.g., do we need `downlevelIteration`?) (**edit**: nah let's do this further down the line, we're already introducing a lot of new breaking changes as is)
- [x] Move some of these `.cjs` files over to `.js`
- [x] Unable to build executable for usage in Docker/GitHub Actions

</details>

## 🧬 QA & Testing

Our existing end-to-end tests GitHub Actions tests are passing, meaning this _should_ work, but I'll do some more QA once we tag a beta release.

You can confirm this all works locally by doing the following:

```sh
# install/build
npm ci && npm run build:exe

# run the JS output (used by npm-based CLI installations)
bin/rdme.js openapi:validate # or whatever command you want

# run the binary
./exe/rdme openapi:validate  # or whatever command you want
```


kanadgupta:
Update: while I made substantial progress on this, we've hit a pretty massive blocker 😵‍💫 we rely on `pkg` for our Docker/GitHub Action build, and turns out it doesn't support ESM at all: https://github.com/vercel/pkg/issues/1291

Jon's work in #857 should make this transition a lot easier, but until we figure out a different strategy for building an executable (or if `pkg` somehow adds ESM support), we probably have to stick with CommonJS for the time being 😭

src/lib/streamSpecToRegistry.ts
```diff
@@ -24,7 +24,14 @@ export default async function streamSpecToRegistry(spec: string) {
...
   const formData = new FormData();
-  formData.append('spec', stream);
+  formData.append('spec', {
+    type: 'application/json',
+    name: 'openapi.json',
+    [Symbol.toStringTag]: 'File',
+    stream() {
+      return stream;
+    },
+  });
```
kanadgupta, line 34: @erunion I'd like your blessing with these changes now that we're using `formdata-node` — does our API care what we name the file? I just set it to `openapi.json` but let me know if you feel otherwise.

.github/workflows/ci.yml
```diff
@@ -15,8 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 14
-          - 16
```
kanadgupta, line 19: with this PR, we're technically requiring Node 18 and above but I'll make that official in a separate PR.

.eslintignore
```diff
@@ -1,3 +1,4 @@
 coverage/
 dist/
-!.alexrc.js
+exe/
+rdme-rollup.cjs
```
erunion, line 4: Why's this config excluded from ESLint?

src/lib/streamSpecToRegistry.ts
```diff
@@ -24,7 +24,14 @@ export default async function streamSpecToRegistry(spec: string) {
 
   debug('file and stream created, streaming into form data payload');
   const formData = new FormData();
-  formData.append('spec', stream);
+  formData.append('spec', {
```
erunion, line 27: This work seems fine but `formdata-node` is still pretty shitty. Recommend moving off it and `node-fetch` as soon as we can. Left a diff in https://github.com/readmeio/rdme/issues/801#issuecomment-1720097752 with the work you can do to this file to move it over to all native APIs.

package.json
```diff
@@ -86,24 +90,27 @@
     "ajv": "^8.11.0",
     "alex": "^11.0.0",
     "eslint": "^8.47.0",
-    "headers-polyfill": "^3.1.2",
+    "headers-polyfill": "^4.0.1",
```
erunion, line 93: Should be able to get rid of this when you move to native `fetch` as well well.

erunion: One thing to not about using native `fetch` is that you'll need to load in `DOM` and `DOM.Iterable` in the tsconfig. DOM for `fetch` and `DOM.Iterable` for `Headers.entries()`.

rollup.config.js
```diff
@@ -0,0 +1,30 @@
...
+    commonjs(),
+    json(),
+    nodeResolve({
+      // see here: https://github.com/rollup/plugins/tree/master/packages/node-resolve#exportconditions
+      // this is required to get chalk working properly
+      exportConditions: ['node'],
+      preferBuiltins: true,
+    }),
+    // this disgusting workaround is required to prevent runtime errors,
+    // see https://github.com/JS-DevTools/ono/issues/19
```
erunion, line 20: I really want to refactor `ono` out of `@readme/openapi-parser` at some point.

src/lib/readmeAPIFetch.ts
```diff
@@ -220,7 +220,9 @@ async function handleRes(res: Response, rejectOnJsonError = true) {
   const contentType = res.headers.get('content-type');
   const extension = mime.extension(contentType);
   if (extension === 'json') {
-    const body = await res.json();
+    // TODO: type this better
```
erunion, line 223: Leaving this as `any` is probably fine imo. You aren't going to have full documented types for these responses without utilizing our API SDK or anything like that (and even then the types are nonexistent for API v1).

erunion (approved):
made it look easy

src/lib/readmeAPIFetch.ts
kanadgupta: yeah `node-fetch` is remarkably bad about typing responses

package.json
erunion: actually, do you need this library? you're loading `Headers` from `node-fetch` in a couple spots already and that seems to be fine.

kanadgupta: yeah i'm thinking i'll grab the types piecemeal using this approach: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924#issuecomment-1707342243

kanadgupta: let me play around, this was technically already an `msw` dep

kanadgupta: okay yeah swapped it out for `node-fetch` for now! see https://github.com/readmeio/rdme/pull/856/commits/5a2a9e33a7b74392da2a664b12db293a1078f732

PR #857, opened 2023-08-10
feat(test): moving unit tests over to vitest

erunion:
## 🧰 Changes

* [x] Move unit tests off Jest and over to [Vitest](https://vitest.dev/).
* ~~Upgrading us to Prettier 3.~~ **(Kanad update 9/5)**: pulled into a separate PR, see https://github.com/readmeio/rdme/pull/864
* ~~Stop testing Node 14 as Prettier 3 doesn't run on it.~~ **(Kanad update 9/5)**: got it working by splitting out the linting into a separate CI step, see https://github.com/readmeio/rdme/pull/864

## 🧬 QA & Testing

If tests pass we should be good to go! Sadly in splitting out the tests and requiring certain tests to be single-threaded, the performance benefits are mostly a wash.

<details>

<summary><strong>🐳 Outdated Performance Benchmarks</strong></summary>

#### Before
```
Test Suites: 33 passed, 33 total
Tests:       1 todo, 502 passed, 503 total
Snapshots:   1 written, 83 passed, 84 total
Time:        81.823 s
Ran all test suites.
npx jest  58.51s user 9.23s system 82% cpu 1:22.60 total
```

#### After
I'm seeing some flakiness in a couple tests but running the entire a bit slower than Jest because we can't use `--threads` due to our usage of `process.chdir()`. See https://github.com/readmeio/rdme/pull/857#discussion_r1289709460 below for some more deets.

```
 Test Files  1 failed | 32 passed (33)
      Tests  1 failed | 501 passed | 1 todo (503)
   Start at  00:39:27
   Duration  126.40s (transform 295ms, setup 110ms, collect 1.66s, tests 124.41s, environment 0ms, prepare 43ms)

npx vitest run  31.68s user 6.17s system 29% cpu 2:06.83 total
```

</details>


vitest.config.ts
```diff
@@ -0,0 +1,27 @@
...
+    setupFiles: ['./__tests__/helpers/vitest.matchers.ts'],
+
+    /**
+     * We can't run tests with `threads` on because we use `process.chrdir()` in some tests and
+     * that isn't available in worker threads, and it's way too much work to mock out an entire
+     * filesystem and `fs` calls for the tests that use it.
+     *
+     * @see {@link https://github.com/vitest-dev/vitest/issues/566}
+     */
+    threads: false,
```
erunion, line 25: It might be possible to refactor the `process.chdir()` stuff away and instead swap in the working directory we're changing into into the paths we're trying to look in? Something like this with `readdirRecursive`:

Screen Shot 2023-08-10 at 12 46 44 AM

If we get rid of `process.chdir()` we should easily be able to shave a minute+ off of the Vitest execution time.

kanadgupta: ah it's used as the `workingDirectory` parameter for resolving refs in our `openapi` family of commands — will look into it!

kanadgupta: sadly we can't move away from the `process.chdir()` stuff yet, we'll need to get rid the `workingDirectory` flag in the `openapi` family of commands. ticketed here: RM-7754

kanadgupta (approved):
@erunion this is your PR but most of the recent work has been mine, mind taking one final glance before I merge?

PR #859, opened 2023-08-15
ci: knip

kanadgupta:
## 🧰 Changes

Trying out [`knip`](https://github.com/webpro/knip) as part of our linting process so we can identify unused code/dependencies. It flagged a few good things!

## 🧬 QA & Testing

Do tests still pass?


src/cmds/categories/create.ts
```diff
@@ -14,7 +14,7 @@ interface Category {
   type: string;
 }
 
-export interface Options {
```
erunion, line 17: I think the reason I exported these originally was in case we wanted to use them in tests but since we aren't doing that, this is a solid find. No sense in exporting something that isn't used internally, and can't/shouldn't be used externally.

package.json
```diff
@@ -84,16 +83,17 @@
     "@types/pluralize": "^0.0.30",
     "@types/prompts": "^2.4.2",
     "@types/semver": "^7.3.12",
-    "@types/tmp": "^0.2.3",
```
erunion, line 87: Did it surface these as unused dependencies?

erunion (approved):

src/cmds/categories/create.ts
kanadgupta: yeah it's great. also a great way to identify unused code, it found some good stuff that we can probably clean out of our `api` subpackage in the main codebase

package.json
kanadgupta: correct!

PR #897, opened 2023-09-14
chore(temp): see if this action paradigm works

kanadgupta:
We currently build a single executable and publish a Docker image to the GitHub container registry, but now that we're using Rollup to build a single JS file, is it possible to use a JavaScript action instead? This PR is a little proof-of-concept of what that would look like.

For reference: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-javascript-actions


kanadgupta:
Update: this was stupidly easy to get working, but gonna close this out for now and revisit in the future. A few downsides:
- it will require a few substantial changes to our release process (on top of everything else v9-related)
- there are a few downsides (losing download count visibility, having to maintain a `dist/` file, etc.)

PR #901, opened 2023-09-14
fix: hide `ExperimentalWarning`

kanadgupta:
## 🧰 Changes

As I feared, the current version in `next` yields this whenever you run an `rdme` CLI command:

CleanShot 2023-09-14 at 18 06 07@2x

This was surprisingly not straightforward (it was _**buried**_ in the comments: https://github.com/nodejs/node/issues/10802#issuecomment-573376999) and I don't love this solution, but I confirmed that adding a `NODE_OPTIONS` env variable to the entry shebang should fix this.

## 🧬 QA & Testing

I `npm link`'d locally and confirmed the warning goes away:

CleanShot 2023-09-14 at 18 09 24@2x

### DIFF
--- a/bin/rdme.js
+++ b/bin/rdme.js
@@ -1,3 +1,5 @@
-#!/usr/bin/env node
+#!/usr/bin/env NODE_OPTIONS=--no-warnings node
+// ^ we need this env variable above to hide the ExperimentalWarnings
+// source: https://github.com/nodejs/node/issues/10802#issuecomment-573376999
 // eslint-disable-next-line import/extensions
 import '../dist/src/cli.js';


bin/rdme.js
```diff
@@ -1,3 +1,5 @@
-#!/usr/bin/env node
+#!/usr/bin/env NODE_OPTIONS=--no-warnings node
```
erunion, line 1: Does that other `process.env.NODE_NO_WARNINGS = 1;` solution work too? I'm not sure which one is cleaner.

erunion (approved):

bin/rdme.js
kanadgupta: nope, doesn't work 😕

PR #903, opened 2023-09-14
chore: type fixes for tests

kanadgupta:
## 🧰 Changes

We generally don't type-check our tests but this PR just silences a bunch of annoying errors when opening up files in my editor.

Technically there's one error left but I'm not sure how to fix it without re-enabling strict mode in the tests folder and dealing with over 200 errors. Again we don't type-check so I'm not pressed about it.

## 🧬 QA & Testing

Do tests still pass?


__tests__/cmds/open.test.ts
```diff
@@ -65,7 +65,7 @@ describe('rdme open', () => {
         return Promise.resolve();
       }
 
-      await expect(cmd.run({ mockOpen, dash: true, key: '12345' })).resolves.toBe(
```
kanadgupta, line 68: wow it actually caught something!

erunion (approved):

PR #905, opened 2023-09-15
chore: tsconfig cleanup

kanadgupta:
## 🧰 Changes

Cleans up and modernizes our `tsconfig`!

## 🧬 QA & Testing

If the build passes across the board we should be good to go.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,20 +2,14 @@
   "compilerOptions": {
     "allowJs": true,
     "baseUrl": "./src",
-    "downlevelIteration": true,
-    "esModuleInterop": true,
     "experimentalDecorators": true,
-    "forceConsistentCasingInFileNames": true,
-    "lib": ["ES2020", "ES2021.Intl"],
+    "lib": ["ES2022"],
     "module": "NodeNext",
     "outDir": "dist/",
-    "paths": {
-      "editor": [".sink.d.ts"]
-    },
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ES2020",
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -2,20 +2,14 @@
   "compilerOptions": {
     "allowJs": true,
     "baseUrl": "./src",
-    "downlevelIteration": true,
-    "esModuleInterop": true,
```
kanadgupta, line 6: this flag is enabled by default now that we're on `NodeNext`

tsconfig.json
```diff
@@ -2,20 +2,14 @@
...
     "experimentalDecorators": true,
-    "forceConsistentCasingInFileNames": true,
```
kanadgupta, line 8: this flag is enabled by default

tsconfig.json
```diff
@@ -2,20 +2,14 @@
...
-    "lib": ["ES2020", "ES2021.Intl"],
+    "lib": ["ES2022"],
     "module": "NodeNext",
     "outDir": "dist/",
-    "paths": {
-      "editor": [".sink.d.ts"]
-    },
```
kanadgupta, line 14: whoops

tsconfig.json
```diff
@@ -2,20 +2,14 @@
...
     "resolveJsonModule": true,
```
erunion, line 9: Are we importing JSON files anywhere not in a unit test?

erunion (approved):

tsconfig.json
kanadgupta: yeah we use our `package.json` file in a bunch of places

PR #906, opened 2023-09-15
feat(versions): flag parity with API, copy fixes

kanadgupta:
## 🧰 Changes

This PR updates our `versions:create` and `versions:update` commands to bring parity with our CLI flags and [our API flags](https://docs.readme.com/main/reference/createversion). The only flag that needed to be updated was `isPublic`, which has been renamed to `hidden`. This PR is a breaking change since the flag needs to be renamed and flipped..

This PR also cleans up our copy related to these commands and cleans up a bunch of tests and comments.

## 🧬 QA & Testing

Does the copy look good and do tests pass?

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,7 +395,7 @@ If you wish to automate the process of creating a new project version, and not h
 For example, the following command contains all the flags to bypass the CLI prompts:
 
 ```sh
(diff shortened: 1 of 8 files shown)


erunion (approved):
oh thank god, camelCase command line arguments always got on my nerves

PR #950, opened 2023-12-04
feat: support `README_API_KEY` env var

### LINKED ISSUE #948: Suggestion: also support `README_API_KEY` in addition to `RDME_API_KEY`
Hi folks 👋 

We recently switched from using the GitHub action to a shell script, because we wanted some extra automation around the deployment. Locally we use `rdme login`, but in CI we just use an API key.

Anyway, I exported the environment variable `README_API_KEY` based on the secret we'd set for the action - `${{ secrets.README_API_KEY }}` - but my command failed telling me I needed a `--key` argument. It was only after re-reading the docs I noticed the subtle difference and I should have used `RDME_...`!

Do you think it might be worth also checking for `README_API_KEY` in addition to the existing one? Seems unlikely I would be using `README_API_KEY` to do something else, so this could be a neat ergonomics improvement IMO.

kanadgupta:
| 🚥 Resolves #948 |
| :------------------- |

## 🧰 Changes

Adds the ability to pass in the API key via the `README_API_KEY` environmental variable.

## 🧬 QA & Testing

Do the docs + tests + etc. look good?

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,10 +103,11 @@ For local CLI usage with a single project, you can authenticate `rdme` to your R
 > [!WARNING]
 > For security reasons, we strongly recommend providing a project API key via the `--key` option in automations or CI environments (GitHub Actions, CircleCI, Travis CI, etc.). It's also recommended if you're working with multiple ReadMe projects to avoid accidentally overwriting existing data.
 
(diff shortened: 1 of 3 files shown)


domharrington (approved):
LGTM!

PR #972, opened 2024-02-07
test: mock out git

kellyjosephprice:
## 🧰 Changes

Speeds up a few tests by mocking out `git` commands.

I was poking around to investigate a bug (https://linear.app/readme-io/issue/RM-8336/childdocs-are-createdsyncd-before-parentdocs-in-rdme-cli) and noticed some test timeouts when running them locally. It looks like for a number of tests, we were making actual requests like `git remote`. I don't understand why those can be so slow, they're nearly instant when running them in a shell?

**my machine before**
image

**my machine after**
image

With all the parallelization, and the fact that the slowest part seems to be the e2e tests, the actual speed up on CI is negligible. But for running individual test files, they go from 5+ seconds to <1 second.

## 🧬 QA & Testing

Tests should pass???

### DIFF
--- a/src/lib/createGHA/index.ts
+++ b/src/lib/createGHA/index.ts
@@ -156,8 +156,8 @@ export default async function createGHA(
 
   // if in a CI environment,
   // don't even bother running the git commands
-  if (!opts.github && (isCI() || isNpmScript())) {
-    debug('not running GHA onboarding workflow in CI and/or npm script, exiting');
+  if (!opts.github && (isCI() || isNpmScript() || (isTest() && !process.env.TEST_RDME_CREATEGHA))) {
+    debug('not running GHA onboarding workflow in CI, npm script, or default test env, exiting 👋');
     return msg;
   }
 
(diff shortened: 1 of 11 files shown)


erunion (approved):
seems ok to me given everything is passing but will let @kanadgupta give the final ✅ before merging

kanadgupta (changes requested):
oh dang, this is a great catch! i have an idea on how to approach this that _should_ be a one or two line change: what if you grabbed [this logic](https://github.com/readmeio/rdme/blob/58bcc4806197ee176dcf3e232d9ad7a9a31e4d33/src/lib/createGHA/index.ts#L181) and moved it up to [this if-statement](https://github.com/readmeio/rdme/blob/58bcc4806197ee176dcf3e232d9ad7a9a31e4d33/src/lib/createGHA/index.ts#L159-L162)? i think you should be able to revert most of the logic you wrote here!

src/lib/createGHA/index.ts
```diff
@@ -156,8 +156,8 @@ export default async function createGHA(
 
   // if in a CI environment,
   // don't even bother running the git commands
-  if (!opts.github && (isCI() || isNpmScript())) {
-    debug('not running GHA onboarding workflow in CI and/or npm script, exiting');
+  if (!opts.github && (isCI() || isNpmScript() || (isTest() && !process.env.TEST_RDME_CREATEGHA))) {
+    debug('not running GHA onboarding workflow in CI, npm script, or default test env, exiting 👋');
```
kanadgupta, line 160: 👋🏽

kanadgupta (approved):
amazing, LGTM! are you still seeing the perf benefits?

kellyjosephprice:
> amazing, LGTM! are you still seeing the perf benefits?

Yea, it looked like the exact same time for test runs.

PR #973, opened 2024-02-07
feat: sort when syncing by parent

kellyjosephprice:
| 🚥 Resolves RM-8336 |
| :------------------ |

## 🧰 Changes

When syncing docs, it sorts them by their parents!

Users reported issues where if a parent document was in a directory, it could error out because it would get synced after one of its children.

## 🧬 QA & Testing

Not sure :grimacing:.

### DIFF
--- a/src/lib/syncDocsPath.ts
+++ b/src/lib/syncDocsPath.ts
@@ -150,7 +150,7 @@ function sortFiles(files: string[], { allowedFileExtensions }: { allowedFileExte
       {},
     );
 
-  const dependencies = Object.values(filesBySlug).reduce(
+  const dependencies = Object.values(filesBySlug).reduce<[string, string][]>(
     (edges, obj) => {
       if (obj.parentDocSlug) {
         edges.push([filesBySlug[obj.parentDocSlug].filePath, filesBySlug[obj.slug].filePath]);
@@ -158,7 +158,7 @@ function sortFiles(files: string[], { allowedFileExtensions }: { allowedFileExte
 
       return edges;
     },
-    [] as [string, string][],
+    [],
   );
 
   return toposort.array(files, dependencies);
--- a/src/lib/syncDocsPath.ts
+++ b/src/lib/syncDocsPath.ts
@@ -141,13 +141,13 @@ function sortFiles(files: string[], { allowedFileExtensions }: { allowedFileExte
 
       return { filePath, slug, parentDocSlug };
     })
-    .reduce(
+    .reduce<Record<string, { filePath: string; parentDocSlug: string; slug: string }>>(
       (bySlug, obj) => {
         // eslint-disable-next-line no-param-reassign
         bySlug[obj.slug] = obj;
         return bySlug;
       },
-      {} as Record<string, { filePath: string; parentDocSlug: string; slug: string }>,
+      {},
     );
 
   const dependencies = Object.values(filesBySlug).reduce(
--- a/src/lib/syncDocsPath.ts
+++ b/src/lib/syncDocsPath.ts
@@ -1,9 +1,9 @@
-import fsSync from 'node:fs';
+import type { ReadDocMetadata } from './readDoc.js';
+
 import fs from 'node:fs/promises';
 import path from 'node:path';
 
 import chalk from 'chalk';
-import frontMatter from 'gray-matter';
 import { Headers } from 'node-fetch';
 import toposort from 'toposort';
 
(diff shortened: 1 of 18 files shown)


src/lib/syncDocsPath.ts
```diff
@@ -127,6 +130,40 @@ async function pushDoc(
...
+      return { filePath, slug, parentDocSlug };
+    })
+    .reduce(
+      (bySlug, obj) => {
+        // eslint-disable-next-line no-param-reassign
+        bySlug[obj.slug] = obj;
+        return bySlug;
+      },
+      {} as Record<string, { filePath: string; parentDocSlug: string; slug: string }>,
+    );
```
kanadgupta, line 151: (forgive me if this is slightly off from a syntax/formatting standpoint) would something like this work?

```suggestion
    .reduce<Record<string, { filePath: string; parentDocSlug: string; slug: string }>>(
      (bySlug, obj) => {
        // eslint-disable-next-line no-param-reassign
        bySlug[obj.slug] = obj;
        return bySlug;
      },
      {},
    );
```

src/lib/syncDocsPath.ts
```diff
@@ -127,6 +130,40 @@ async function pushDoc(
...
+  const dependencies = Object.values(filesBySlug).reduce(
+    (edges, obj) => {
+      if (obj.parentDocSlug) {
+        edges.push([filesBySlug[obj.parentDocSlug].filePath, filesBySlug[obj.slug].filePath]);
+      }
+
+      return edges;
+    },
+    [] as [string, string][],
+  );
```
kanadgupta, line 162: similar to above:

```suggestion
  const dependencies = Object.values(filesBySlug).reduce<[string, string][]>(
    (edges, obj) => {
      if (obj.parentDocSlug) {
        edges.push([filesBySlug[obj.parentDocSlug].filePath, filesBySlug[obj.slug].filePath]);
      }

      return edges;
    },
    [],
  );
```

src/lib/syncDocsPath.ts
```diff
@@ -127,6 +130,40 @@ async function pushDoc(
...
+function sortFiles(files: string[], { allowedFileExtensions }: { allowedFileExtensions: string[] }): string[] {
+  const extensionRegexp = new RegExp(`(${allowedFileExtensions.join('|')})$`);
+
+  const filesBySlug = files
+    .map(filePath => {
+      const doc = frontMatter(fsSync.readFileSync(filePath, 'utf8'));
+      const slug = path.basename(filePath).replace(extensionRegexp, '');
+      const parentDocSlug = doc.data.parentDocSlug;
+
+      return { filePath, slug, parentDocSlug };
```
kanadgupta, line 142: we have a `readDoc` utility that should take care of this, mind using that instead? also it appears we're reading each file twice now, any way we could optimize that[^1]?

[^1]: just flagging it so we're aware, but i don't think it's a real issue!

src/lib/syncDocsPath.ts
```diff
@@ -127,6 +130,40 @@ async function pushDoc(
...
+        // eslint-disable-next-line no-param-reassign
+        bySlug[obj.slug] = obj;
+        return bySlug;
+      },
+      {} as Record<string, { filePath: string; parentDocSlug: string; slug: string }>,
+    );
+
+  const dependencies = Object.values(filesBySlug).reduce(
+    (edges, obj) => {
+      if (obj.parentDocSlug) {
```
kanadgupta, line 155: i know we have users that sync only parts of their docs using `rdme`, i think we should probably also ensure that the parent file actually exists

(also let me know if i'm misunderstanding how this works!)

```suggestion
      if (obj.parentDocSlug && filesBySlug[obj.parentDocSlug]) {
```

src/lib/syncDocsPath.ts
```diff
@@ -127,6 +130,40 @@ async function pushDoc(
     });
 }
 
+function sortFiles(files: string[], { allowedFileExtensions }: { allowedFileExtensions: string[] }): string[] {
```
kanadgupta, line 133: this function appears to use `parentDocSlug`, is there any way we add basic support for `parentDoc`? i know it uses object IDs so there's no way for us to build a graph, but at the very least we could make it so docs that have a `parentDoc` attribute get synced after ones that lack one?

kanadgupta (changes requested):
oh wow this is a very sophisticated approach, love it! i have some feedback + questions below, happy to pair on any of this!

src/lib/syncDocsPath.ts
```diff
@@ -27,10 +30,10 @@ async function pushDoc(
   key: string,
   selectedVersion: string | undefined,
   dryRun: boolean,
-  filePath: string,
+  fileData: ReadDocMetadata,
   type: CommandCategories,
 ) {
-  const { content, data, hash, slug } = readDoc(filePath);
+  const { content, data, filePath, hash, slug } = fileData;
```
kanadgupta, line 36: this refactor (i.e., decoupling the `readDoc` so we can read the contents prior to sorting) is very clever!

kanadgupta (approved):
image

PR #993, opened 2024-04-24
fix: use newest import attributes syntax

kanadgupta:
## 🧰 Changes

TC39 updated their import _assertions_ to be import _attributes_ and it's a bit of a mess now that we're running tests against Node v22.

Here's what importing JSON in ESM used to look like:

```js
import { x } from "./mod" assert { type: "json" }
```

now it looks like this:

```js
import { x } from "./mod" with { type: "json" };
```

Support for this new `with` keyword was added in [Node v18.20.0](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#added-support-for-import-attributes) and [Node v20.10.0](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#2023-11-22-version-20100-iron-lts-targos) (and is supported on all v21 and v22 channels), but importantly the older `assert` keyword is **_not_** supported in Node v22, so unfortunately our `package.json#engines.node` field is going to look a little ridiculous for the time being.

Details: https://github.com/tc39/proposal-import-attributes?tab=readme-ov-file#history

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "author": "ReadMe <support@readme.io> (https://readme.com)",
   "engines": {
-    "node": ">=18"
+    "node": ">=18.20.0"
   },
   "bin": {
     "rdme": "bin/rdme.js"
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "author": "ReadMe <support@readme.io> (https://readme.com)",
   "engines": {
-    "node": ">=18.20.0"
+    "node": "^18.20.0 || >=20.10.0"
   },
   "bin": {
     "rdme": "bin/rdme.js"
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "author": "ReadMe <support@readme.io> (https://readme.com)",
   "engines": {
-    "node": ">=18"
+    "node": "^18.20.0 || >=20.10.0"
```
erunion, line 8: Do our compiled dists contain these import assertion so? If not then this bump is really just for building it?

kanadgupta: they do yeah 😔 all these files contain the import attributes and make it into the compiled dist:

erunion (approved):

package.json
kanadgupta: here's an example dist file:

CleanShot 2024-04-24 at 17 39 53@2x

kanadgupta: yep — it's just a single CLI dist so no `tsup` madness needed here thankfully

PR #1063, opened 2024-11-11
Oleh/openapi adding command to solve circularity and recursiveness

olehshh:
| 🚥 Resolves RM-11373 |
| :------------------- |

## 🧰 Changes

1. Adding a script to automate the workaround for circular and recursive references.
2. Adding this command to the list of existing ones.

## 🧬 QA & Testing

Take any file with circular references, for example
```
openapi: 3.0.0
info:
  title: Cyclic Reference Example
  version: 1.0.0
paths:
  /example:
    get:
      summary: Example endpoint with cyclic references
      responses:
        '200':
          description: Successful response
components:
  Schemas:
    ObjectC:
      type: object
      properties:
        test:
          type: string
    ObjectA:
      type: object
      properties:
        test:
          type: string
        relatedObject:
          $ref: '#/components/schemas/ObjectB'
    ObjectB:
      type: object
      properties:
        test:
          type: string
        relatedObject:
          type: array
          items:
            $ref: '#/components/schemas/ObjectA'
        test2:
          $ref: '#/components/schemas/ObjectC' 
```

Execute the `rdme openapi:refs [path]` command and use `inspect` to check for circular references after execution. The command also handles recursiveness.

### DIFF
--- a/src/cmds/openapi/refs.ts
+++ b/src/cmds/openapi/refs.ts
@@ -1,0 +1,221 @@
+/* eslint-disable no-param-reassign */
+import type { ZeroAuthCommandOptions } from '../../lib/baseCommand.js';
+import type { OASDocument } from 'oas/types';
+import type { IJsonSchema } from 'openapi-types';
+
+import fs from 'node:fs';
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import jsYaml from 'js-yaml';
+
+import analyzeOas from '../../lib/analyzeOas.js';
+import Command, { CommandCategories } from '../../lib/baseCommand.js';
+
+interface Options {
+  spec?: string;
+}
+
+type SchemaCollection = Record<string, IJsonSchema>;
+
+class OpenAPISolvingCircularityAndRecursiveness extends Command {
+  constructor() {
+    super();
+    this.command = 'openapi:refs';
+    this.usage = 'openapi:refs [file]';
+    this.description = 'The script resolves circular and recursive references in OpenAPI by replacing them with object schemas. However, not all circular references can be resolved. You can run the openapi:inspect command to identify which references remain unresolved.';
+    this.cmdCategory = CommandCategories.APIS;
+
+    this.hiddenArgs = ['spec'];
+    this.args = [
+      {
+        name: 'spec',
+        type: String,
+        defaultOption: true,
+      },
+    ];
+  }
+
+  /**
+   * Reads and parses an OpenAPI file (JSON or YAML).
+   * @param {string} filePath - The file path to read.
+   * @returns {any} The parsed content of the file.
+   */
+  static readOpenApiFile(filePath: string) {
+    const fileContent = fs.readFileSync(filePath, 'utf8');
+    return filePath.endsWith('.json') ? JSON.parse(fileContent) : jsYaml.load(fileContent);
+  }
+
+  /**
+   * Writes OpenAPI data to a file (JSON or YAML).
+   * @param {string} filePath - The file path to write to.
+   * @param {OASDocument} data - The data to be written.
+   */
+  static writeOpenApiFile(filePath: string, data: OASDocument) {
+    const content = filePath.endsWith('.json')
+      ? JSON.stringify(data, null, 2)
+      : jsYaml.dump(data, { noRefs: true }); // Disables YAML anchors
+    fs.writeFileSync(filePath, content, 'utf8');
+  }
+
+  /**
+   * Identifies circular references in the OpenAPI document.
+   * @param {OASDocument} document - The OpenAPI document to analyze.
+   * @returns {Promise<string[]>} A list of circular reference paths.
+   */
+  static async getCircularRefsFromOas(document: OASDocument): Promise<string[]> {
+    try {
+      const analysis = await analyzeOas(document);
+      const circularRefs = analysis.openapi.circularRefs;
+      return Array.isArray(circularRefs.locations) ? circularRefs.locations : [];
+    } catch (error) {
+      return [`Error analyzing OpenAPI document: ${error}`];
+    }
+  }
+
+  /**
+   * Replaces a reference in a schema with an object if it's circular or recursive.
+   * @param {IJsonSchema} schema - The schema to process.
+   * @param {string[]} circularRefs - List of circular reference paths.
+   * @param {string} schemaName - The name of the schema being processed.
+   * @returns {IJsonSchema} The modified schema or the original.
+   */
+  static replaceRefWithObject(schema: IJsonSchema, circularRefs: string[], schemaName: string): IJsonSchema {
+    if (schema.$ref) {
+      const refSchemaName = schema.$ref.split('/').pop() as string;
+      const isCircular = circularRefs.some((refPath) => refPath.includes(refSchemaName));
+      const isRecursive = schemaName === refSchemaName;
+
+      if (schemaName.includes('Ref') && (isCircular || isRecursive)) {
+        return { type: 'object' } as IJsonSchema;
+      }
+    }
+    
+    return schema;
+  }
+
+  /**
+   * Recursively replaces references in schemas, transforming circular references to objects.
+   * @param {IJsonSchema} schema - The schema to process.
+   * @param {string[]} circularRefs - List of circular reference paths.
+   * @param {string} schemaName - The name of the schema being processed.
+   */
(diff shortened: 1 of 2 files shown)


src/cmds/openapi/refs.ts
```diff
@@ -0,0 +1,235 @@
...
+   * @returns {Promise<string>} Result message.
+   */
+  async run(opts: ZeroAuthCommandOptions<Options>): Promise<string> {
+    await super.run(opts);
+    const { spec } = opts;
+    if (!spec) {
+      return 'File path is required.';
+    }
+
+    const openApiData = OpenAPISolvingCircularityAndRecursiveness.readOpenApiFile(spec);
```
kanadgupta, line 203: rather than using this static method to read the file, can you use our existing `prepareOas` function? see an example here: https://github.com/readmeio/rdme/blob/5db25a1b5bec9c26966eda5cd1d616a4de668601/src/cmds/openapi/convert.ts#L57

src/cmds/openapi/refs.ts
```diff
@@ -0,0 +1,235 @@
...
+        remainingCircularRefs = remainingCircularRefs.length > 0 ? [] : remainingCircularRefs;
+        iterationCount += 1;
+      }
+
+      if (iterationCount >= 5) {
+        return 'Maximum iteration limit reached. Some circular references may remain unresolved.';
+      }
+    }
+
+    OpenAPISolvingCircularityAndRecursiveness.writeOpenApiFile(spec, openApiData);
```
kanadgupta, line 230: rather than overwriting the existing file, can you add a prompt (and a corresponding `--outputPath` flag) so users can specify what the updated file path should be? see here for an example:

https://github.com/readmeio/rdme/blob/5db25a1b5bec9c26966eda5cd1d616a4de668601/src/cmds/openapi/convert.ts#L66-L85

kanadgupta (changes requested):
looks great so far! haven't had a chance to properly test this yet, but some preliminary feedback below! also if you could add some docs in `README.md` that'd be great. thanks!

kanadgupta (changes requested):
DM'd you about this but this has a few merge conflicts now with the changes that were shipped in #1068 — happy to help with these if you need a hand!

olehshh:
A provisional approach to prevent modifications to files with unresolved references that cannot currently be addressed.

src/index.ts
```diff
@@ -11,6 +11,7 @@ import OpenAPIConvertCommand from './commands/openapi/convert.js';
 import OpenAPICommand from './commands/openapi/index.js';
 import OpenAPIInspectCommand from './commands/openapi/inspect.js';
 import OpenAPIReduceCommand from './commands/openapi/reduce.js';
+import OpenAPISolvingCircularityAndRecursiveness from './commands/openapi/refs.js';
```
kanadgupta, line 14:
```suggestion
import OpenAPIRefsCommand from './commands/openapi/refs.js';
```

src/commands/openapi/refs.ts
```diff
@@ -0,0 +1,323 @@
...
+import BaseCommand from '../../lib/baseCommand.js';
+import { titleFlag, workingDirectoryFlag } from '../../lib/flags.js';
+import { warn, debug } from '../../lib/logger.js';
+import prepareOas from '../../lib/prepareOas.js';
+import promptTerminal from '../../lib/promptWrapper.js';
+import { validateFilePath } from '../../lib/validatePromptInput.js';
+
+type SchemaCollection = Record<string, IJsonSchema>;
+
+export default class OpenAPISolvingCircularityAndRecursiveness extends BaseCommand<
```
kanadgupta, line 22:
```suggestion
export default class OpenAPIRefsCommand extends BaseCommand<
```

src/commands/openapi/refs.ts
```diff
@@ -0,0 +1,323 @@
...
+  typeof OpenAPISolvingCircularityAndRecursiveness
```
kanadgupta, line 23:
```suggestion
  typeof OpenAPIRefsCommand
```

src/index.ts
```diff
@@ -45,6 +46,7 @@ export const COMMANDS = {
   'openapi:inspect': OpenAPIInspectCommand,
   'openapi:reduce': OpenAPIReduceCommand,
   'openapi:validate': OpenAPIValidateCommand,
+  'openapi:refs': OpenAPISolvingCircularityAndRecursiveness,
```
kanadgupta, line 49: alphabetizing and renaming:

```suggestion
  'openapi:refs': OpenAPIRefsCommand,
  'openapi:validate': OpenAPIValidateCommand,
```

src/commands/openapi/refs.ts
```diff
@@ -0,0 +1,323 @@
...
+> {
+  static description =
+    'Resolves circular and recursive references in OpenAPI by replacing them with object schemas. Not all circular references can be resolved automatically.';
```
kanadgupta, line 26: can you add a `summary` field to this class that contains just this part of this description:

> Resolves circular and recursive references in OpenAPI by replacing them with object schemas.

src/commands/openapi/refs.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  static args = {
+    spec: Args.string({ description: 'A file/URL to your API definition' }),
+  };
+
+  static flags = {
+    out: Flags.string({ description: 'Output file path to write processed file to' }),
+    title: titleFlag,
```
kanadgupta, line 34: i don't believe this flag is used:

```suggestion
```

documentation/commands/openapi.md
```diff
@@ -257,3 +258,39 @@ EXAMPLES
 
     $ rdme openapi:validate
 ```
+
+## `rdme openapi:refs [SPEC]`
```
kanadgupta, line 262: this document is auto-generated from the code — you can regenerate this by running `npm run build && npm run build:docs`. can you do that instead? the docs you wrote here are great and we should definitely use them! you should be able to supply them via the `summary`, `description`, and `example` methods on your main command class.

__tests__/__snapshots__/index.test.ts.snap
kanadgupta, line 1: can you delete this file?

kanadgupta (changes requested):
thank you for your patience with all of the work we've been doing in this repo 😮‍💨 it's very close to the finish line!

__tests__/lib/__snapshots__/analyzeOas.test.ts.snap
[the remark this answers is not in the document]
olehshh: I returned this file, there was some confusion with them(

__tests__/__fixtures__/circular-references-oas/circular-reference-ref-only-resolved.json
```diff
@@ -0,0 +1,31 @@
...
+      "ObjectA": {
+        "type": "object",
+        "properties": {
+          "referenceToB": {
+            "$ref": "#/components/schemas/ObjectARef"
+          }
+        }
+      },
+      "ObjectB": {
+        "$ref": "#/components/schemas/ObjectA"
```
olehshh, line 19: given that this schema contains nothing but a reference to objectA, objectA can be considered recursive and this result does not change the functionality, only the visual

__tests__/__fixtures__/circular-references-oas/circular-reference-ref-only-title-resolved.json
```diff
@@ -0,0 +1,32 @@
...
+      "ObjectA": {
+        "type": "object",
+        "properties": {
+          "referenceToB": {
+            "$ref": "#/components/schemas/ObjectB"
+          }
+        }
+      },
+      "ObjectB": {
+        "title": "Cyclic Schema Example",
```
olehshh, line 19: instead, if the schema contains more than just a reference, the handler treats it as a separate schema accordingly

erunion:
Okay @emilyskuo and @kanadgupta this is ready for review. @olehshh had this in pretty solid shape already I just slightly reworked a couple things:

* Renamed the command from `openapi refs` to `openapi resolve`
* Rewrote the test suite to be more akin to how our other `openapi` command tests look and added another test for a failure case.
* Changed the static methods in the command to be `private` and replaced their usages of `debug()`, `warn()`, and `info()` with `this.*` calls. 
* Added the ability to call `this.info()` into our `BaseCommand`.

PR #1068, opened 2024-11-16
feat: oclif (take 2)

kanadgupta:
## 🧰 Changes

i took a stab at this in [#962](https://github.com/readmeio/rdme/pull/962) but ran into issues at the time but oclif has since shipped a lot of handy features (and plus the work that folks did on `owl` got me excited about oclif again) so i decided to revive this and i got it all working 🥹

here's a high-level overview:
- [x] refactors all commands, associated helper functions, and github actions build process to be oclif-friendly
- [x] converts all tests to use oclif test utilities

still outstanding:

- [x] a handful of tests are still failing, hoping that https://github.com/oclif/test/pull/652 gets merged in)
- [x] github actions dry runs are failing (but i got them working in [#1067](https://github.com/readmeio/rdme/pull/1067))[^1]
- [x] remove all docker build/tagging/release code
- [x] update `semantic-release` workflow to support this new process. here are the assets we'll need to consider every time we tag a release ([link](https://github.com/readmeio/rdme/blob/d01d76fe3c2e4a98b4f5c415be03e589fbe3b56e/.releaserc.yml#L30)):
  - [x] command list README (i.e., the one generated by `oclif readme`
  - [x] we do **not** need to update `action.yml` on every release anymore
  - [x] everything in `dist-gha` (the two JS files ~~and `package.json`~~) 
    -  [x] (do we need an `oclif` manifest in this directory? probably not?)
    -  [x] we might not actually need to track this duplicated `package.json` — what if we used the [`runs.pre`](https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runspre) script to generate it?
  - [x] package/package-lock.json
  - [x] CHANGELOG.md
  - [x] `oclif` manifest (this one is interesting... we probably shouldn't git track it but it should be included in our npm assets)

## 🧬 QA & Testing

to test this out, check out this branch and play around:

```sh
npm ci
bin/dev.js whoami
```

i've confirmed in a separate repo that the action works as expected: https://github.com/readmeio/oas-examples/actions/runs/11902772744/job/33168462740

i ticketed a few follow-ups in RM-11447

(resolves RM-8260)

[^1]: i'm going to move away from the docker-based github action builds — it's pretty complex and unfortunately doesn't play nicely with oclif. i figured out a way to get it working with pre-built JS dists in [#1067](https://github.com/readmeio/rdme/pull/1067) that is just as fast (if not faster) than before.


__tests__/helpers/setup-oclif-config.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function runCommand<T extends typeof OclifCommand>(Command: T) {
+  return async function runCommandAgainstArgs(args?: string[]) {
+    const oclifConfig = await setupOclifConfig();
+    // @ts-expect-error this is the pattern recommended by the @oclif/test docs.
+    // Not sure how to get this working with type generics.
+    return captureOutput<string>(() => Command.run(args, oclifConfig), { testNodeEnv: 'rdme-test' }).then(
+      ({ error, result }) => {
+        if (error) {
```
erunion, line 33: Sucks that `@oclif/test` seems to clobber the `SyntaxError` errors that `prepareOas` throws.

Screenshot 2024-11-17 at 12 05 42 PM

erunion: Oh looks like you PR'd a fix to them for this? https://github.com/oclif/test/pull/652

__tests__/cmds/openapi/index.test.ts
```diff
@@ -59,6 +60,7 @@ describe('rdme openapi', () => {
   beforeEach(() => {
     consoleInfoSpy = vi.spyOn(console, 'info').mockImplementation(() => {});
     consoleWarnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
```
erunion, line 62: I've always disliked having to do this and as oclif has some tooling in `captureOutput` to capture logs it. might be if we could move away from this somehow. Looks like we use the logging library to handle how logging happens in GHA but maybe those could be encapsulated into our `BaseCommand`?

Might be too big of a lift just for the sake of eliminating some unit test spites though.

__tests__/helpers/setup-oclif-config.ts
```diff
@@ -0,0 +1,42 @@
...
+    root,
+    version: '7.0.0',
+  });
+}
+
+export function runCommand<T extends typeof OclifCommand>(Command: T) {
+  return async function runCommandAgainstArgs(args?: string[]) {
+    const oclifConfig = await setupOclifConfig();
+    // @ts-expect-error this is the pattern recommended by the @oclif/test docs.
+    // Not sure how to get this working with type generics.
```
erunion, line 30: Looked into this a bit and I don't think it's possible for us to do because arguments and flags are supplied differently to the CLI and need actual string parsing to determine what you're doing.

src/cli.ts
```diff
@@ -1,55 +0,0 @@
...
-
-    /**
-     * If we're in a GitHub Actions environment, log errors with that formatting instead.
-     *
-     * @see {@link https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message}
-     * @see {@link https://github.com/actions/toolkit/tree/main/packages/core#annotations}
-     */
-    if (isGHA()) {
-      return core.setFailed(message);
-    }
```
erunion, line 42: Will probably want to retain this somewhere.

package.json
```diff
@@ -141,5 +149,45 @@
...
     }
   },
+  "oclif": {
+    "bin": "rdme",
+    "dirname": "rdme",
+    "commands": {
+      "strategy": "explicit",
+      "target": "./dist/index.js",
+      "identifier": "COMMANDS"
+    },
```
erunion, line 159: Why not let oclif automatically pick up commands?

package.json
```diff
@@ -110,7 +118,7 @@
     "debug": "tsx src/cli.ts",
     "lint": "alex . && knip && npm run lint:ts && npm run prettier && npm run schemas:check",
     "lint:ts": "eslint . --ext .js,.ts",
-    "prebuild": "rm -rf dist/",
+    "prebuild": "rm -rf dist/ && rm -f src/package.json && ln package.json src/package.json",
```
erunion, line 121: Ran into a small problem with this where upon first cloning down the branch I needed to set up this `package.json` link before being able to run `./bin/dev.js`. Might want to add a note about it to the CONTRIBUTING file.

src/cmds/categories/create.ts
```diff
@@ -13,60 +12,34 @@ interface Category {
...
+  static flags = {
+    categoryType: Flags.option({
+      description: 'Category type',
+      options: ['guide', 'reference'] as const,
+      required: true,
+    })(),
+    key: keyFlag,
+    preventDuplicates: Flags.boolean({
+      description:
+        'Prevents the creation of a new category if there is an existing category with a matching `categoryType` and `title`',
```
erunion, line 31: Definitely doesn't need to happen here but before we publish a v9 with APIv2 support we should do a full pass on all flag and argument descriptions, the punctuation and terminology in some of them could use some consistency passes.

src/cmds/open.ts
erunion, line 1: Honestly wonder how much usage this command gets.

__tests__/helpers/setup-oclif-config.ts
kanadgupta: yep exactly

__tests__/cmds/openapi/index.test.ts
kanadgupta: yeah def agreed — i have some follow-ups in my internal notes (that i'll document as proper tickets) to get rid of our `lib/logger` entirely in favor of oclif's logging but the refactor would be pretty involved and blow up the diff even further so i'll try and get that done in a follow-up PR at some point

package.json
kanadgupta: we bundle the `src/index.ts` (i.e., `dist/index.js`) entrypoint for our github action bundles — see the `rollup.config.js` changes in #1067

src/cmds/categories/create.ts
kanadgupta: totally agreed! i'm planning on doing a deep revamp of our docs (including a migration guide) prior to release

src/cmds/open.ts
kanadgupta: probably not much — i'll keep it for now (at least in this PR) since it's not much additional overhead to support

kanadgupta: i also will probably get rid of the `validate` alias in favor of `openapi:validate`

package.json
kanadgupta: ah, that's a great callout ty

__tests__/helpers/setup-oclif-config.ts
kanadgupta: yeah so the goal here was to make it so `npx vitest --watch [command-file].ts` properly reloads every time i make a change to the command file (which is surprisingly challenging b/c `oclif` does a lot of magical command module importing by default). this TS error isn't great but otherwise i'm pretty content with this setup overall (open to ideas though)

src/cmds/changelogs.ts
```diff
@@ -1,47 +1,34 @@
...
-  filePath?: string;
-}
-
-@supportsGHA
-export default class ChangelogsCommand extends Command {
-  constructor() {
-    super();
+export default class ChangelogsCommand extends BaseCommand<typeof ChangelogsCommand> {
+  // we need this as a const for syncDocsPath
+  id = 'changelogs' as const;
```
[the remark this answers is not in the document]
kanadgupta: ugh oh yeah i did the `syncDocsPath` logic ages ago and only just added the unit test paradigm recently so i forgot we're doing both 😭

the way that we pass command data to the `createGHA` hook vs. the way that we invoke `syncDocsPath` is slightly different. definitely open to refactoring this but i don't want to have to do a deep-dive on `@oclif/core` types again rn lol

src/cli.ts
kanadgupta: it's been a bit since i wrote all the `baseCommand` logic but yeah i can look into it. tbh i don't actually think this function is doing much but i can double check to see if the error output is any different:

https://github.com/actions/toolkit/blob/bb2278e5cfbb40afc20890c415e9ffa836631cd5/packages/core/src/core.ts#L220-L224

package.json
kanadgupta: https://github.com/readmeio/rdme/pull/1068/commits/e990a2bbda5ac30b78fba890256d435474853534

__tests__/helpers/setup-oclif-config.ts
kanadgupta: fixed 🥹 https://github.com/readmeio/rdme/pull/1068/commits/6f858aab220552652a46852a85dc1fe58c7b5600

PR #1069, opened 2024-11-17
refactor(oclif): renaming `src/cmds` to `src/commands`

erunion:
## 🧰 Changes

This is a super minor addendum to the oclif work going on in https://github.com/readmeio/rdme/pull/1068 but I've renamed our `src/cmds` directory to `src/commands` as when you run `oclif generate command <command name>` that's the directory it wants to place a new command into.

We could _probably_ configure `src/cmds` somehow to be the place it looks but I don't see anything in their [config docs](https://oclif.io/docs/config) about it.


kanadgupta (approved):
imma wait until #1068 is merged into `next` before merging this in because the diff is already crazy

kanadgupta:
closing in favor of https://github.com/readmeio/rdme/pull/1070

PR #1082, opened 2024-11-22
fix(ci): semantic-release workflow for v9 releases

kanadgupta:
## 🧰 Changes

sets up the `semantic-release` configuration so we can properly maintain v9 releases.

further reading: https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#maintenance-branches

outstanding work:

- [x] make sure that the major tag is only set on the `main` branch release (e.g., if `v10.0.1` is being published then the major tagging step should run, but if we're publishing a `v9.0.1` release then the major tagging step should **not** run)
- [x] **(do this right before merging)** make sure a `v9` branch is created and it's pointing to the same commit as the `v9` tag
- [ ] **(do this right after merging)** if the release was successfully published to the `v9` branch, then delete the `v9` tag

## 🧬 QA & Testing

we truly won't know if this works until it's merged 🫠 but that's why we have a `next` branch right?


mjcuva (approved):
I'm not super familiar with this package, but skimming the docs I think this seems good to me!

PR #1084, opened 2024-11-22
test: rework how we assert command results

kanadgupta:
## 🧰 Changes

this PR is a slight rework to how we assert command results using `@oclif/test` (inspired by https://github.com/oclif/test/issues/655). we use a mishmash of our own internal logger vs `oclif`'s `this.log` and this sets us up for better success when we eventually unify all of this in v10.

this also touches up our vitest setup a bit — hiding Node.js `ExperimentalWarning` outputs, removing an unused matcher, only importing a matcher where we use it, etc.

our vitest output has been a bit of a mess lately so all of these changes culminate in a much cleaner output and i'm irrationally excited about it.

<details><summary>before</summary>

CleanShot 2024-11-22 at 10 55 54@2x

</details>

<details><summary>after</summary>

CleanShot 2024-11-22 at 10 55 21@2x

</details>

## 🧬 QA & Testing

no functional changes nor changes in test coverage — if tests still pass we should be in good shape.


__tests__/commands/changelogs/index.test.ts
```diff
@@ -21,7 +21,13 @@ describe('rdme changelogs', () => {
...
   beforeAll(() => {
     nock.disableNetConnect();
-    run = runCommand(Command);
+    run = (args: string[]) =>
+      runCommand(Command)(args).then(({ error, result }) => {
+        if (error) {
+          throw error;
+        }
+        return result;
+      });
```
erunion, line 30: since these few lines of code are being reused in every test could they be made more generic and moved into `helpers/oclif`?

vitest.config.ts
```diff
@@ -6,6 +6,9 @@ export default defineConfig({
     coverage: {
       exclude: [...coverageConfigDefaults.exclude, '**/dist-gha/**'],
     },
+    // We'll defer to `@oclif/test` for console interception
+    // so we can run assertions against console output.
+    disableConsoleIntercept: true,
```
erunion, line 11: neat

vitest.config.ts
```diff
@@ -19,6 +22,8 @@ export default defineConfig({
        * tool in a testing environment.
        */
       NODE_ENV: 'rdme-test',
+      // Node emits ExperimentalWarnings because we import JSON modules, so this hides that output.
+      NODE_OPTIONS: '--disable-warning=ExperimentalWarning',
```
erunion, line 26: rad that vitest has the ability to configure these

erunion (approved):

__tests__/commands/changelogs/index.test.ts
kanadgupta: https://github.com/readmeio/rdme/pull/1084/commits/f79579383a37487224e0aa6d71b2c91a6f8c097d

erunion (approved):

PR #1085, opened 2024-11-22
feat!: drop node 18

kanadgupta:
## 🧰 Changes

BREAKING CHANGE: dropping support for Node.js v18. The minimum required Node.js version is v20.10.0.


package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "author": "ReadMe <support@readme.io> (https://readme.com)",
   "engines": {
-    "node": "^18.20.0 || >=20.10.0"
+    "node": ">=20.10.0"
```
erunion, line 8: should CI get updated for this so we stop testing LTS-1 until node 24 is out?

erunion (approved):

package.json
kanadgupta: turns out [LTS-1 is v20](https://github.com/readmeio/rdme/actions/runs/12126551205/job/33808878042#step:3:10) since node 22 is now LTS so we haven't been running tests against node 18 for sometime now lol 🫢

erunion: ope

PR #1115, opened 2024-12-09
fix: executable issue for linux users

kanadgupta:
## 🧰 Changes

Linux requires the `-S` flag when passing options to `node` command-line in executable files. This was fun to troubleshoot 🥴

## 🧬 QA & Testing

Can Linux users run the CLI again? 😮‍💨

### DIFF
--- a/bin/dev.js
+++ b/bin/dev.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env npx tsx
+#!/usr/bin/env -S npx tsx
 
 async function main() {
   const { execute } = await import('@oclif/core');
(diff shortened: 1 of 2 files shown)


kellyjosephprice (approved):
TIL: `env -S`, how does that fix this???!

kanadgupta:
@kellyjosephprice i'm flying blind here so i could not tell you lol

joshbeard:
For what it's worth, the `-S` option breaks in environments that use BusyBox (e.g. the `node:alpine` Docker images). A solution there is to install `coreutils`.

```
/usr/bin/env: unrecognized option: S
BusyBox v1.37.0 (2024-11-19 21:09:16 UTC) multi-call binary.

Usage: env [-i0] [-u NAME]... [-] [NAME=VALUE]... [PROG ARGS]

Print current environment or run PROG after setting up environment

	-, -i	Start with empty environment
	-0	NUL terminated output
	-u NAME	Remove variable from environment
```

kanadgupta:
@joshbeard appreciate you flagging this — mind taking a look at https://github.com/readmeio/rdme/pull/1117 and seeing if that takes care of this issue? If not, might be good to add additional docs to our setup instructions.

PR #1117, opened 2024-12-10
fix: remove import attributes

kanadgupta:
## 🧰 Changes

The whole reasoning behind #1115 is because we pass flags in order to avoid emitting this `ExperimentalWarning` for Node v20 users:

CleanShot 2024-12-10 at 12 12 00@2x

Per [this comment](https://github.com/readmeio/rdme/pull/1115#issuecomment-2532123627), adding `-S` might be problematic for certain container setups. As a result, I just overhauled how we import JSON in our production code so we don't need that Node CLI flag at all.

At some point I'd love if we got to a point where we can fully rely on the oclif config for `package.json` attributes but today is not that day!

## 🧬 QA & Testing

When you check out this branch and run the following commands...

- Do they run properly without hanging? (hopefully yes)
- Do you see Node `ExperimentalWarning` outputs (like the above screenshot) at any point? (hopefully no)

```sh
npm ci && npm run build
bin/run.js
```


kellyjosephprice (approved):
Works locally for me

PR #1126, opened 2024-12-21
feat(openapi): tooling upgrades

erunion:
## 🧰 Changes

This upgrades our [oas](https://npm.im/oas) and [oas-normalize](https://npm.im/oas-normalize) tooling to their latest releases. Included in this work are a couple new features to and refactors to some commands:

#### `openapi`

The `.validate()` call within `oas-normalize` has been refactored to no longer do conversion and validation, the conversion side has been split off to a new `.convert()` method.

As part of this change the previous `convertToLatest` parameter that `.validate()` accepted for converting Swagger definitions to OpenAPI no longer exists anywhere. We had this configurable within some internal methods within rdme but it was never exposed, and converting Swagger to OpenAPI has always been the default behavior throughout rdme.

#### `openapi inspect`

Our analyzer tool can now surface information on if an API definition utilizes common parameters.[^1]

I have also removed our dependency on [pluralize](https://npm.im/pluralize) because we were only loading it to pluralize three strings: "media type", "operation", and "security type". Because these can be easily pluralized with a one-liner we don't really need to import a wholeass library to do this.

[^1]: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#path-item-object


src/lib/prepareOas.ts
```diff
@@ -189,12 +183,20 @@ export default async function prepareOas(
...
-  let api: OpenAPI.Document = await oas.validate({ convertToLatest: opts.convertToLatest }).catch((err: Error) => {
+  let api: OpenAPI.Document;
+  await oas.validate().catch((err: Error) => {
     spinner.fail();
     debug(`raw validation error object: ${JSON.stringify(err)}`);
     throw err;
   });
+
+  // If we were supplied a Postman collection this will **always** convert it to OpenAPI 3.0.
+  api = await oas.convert().catch((err: Error) => {
```
kanadgupta, line 194: we shouldn't need to do this if we're validating, correct?

src/lib/prepareOas.ts
```diff
@@ -189,12 +183,20 @@ export default async function prepareOas(
...
-  // If we were supplied a Postman collection this will **always** convert it to OpenAPI 3.0.
-  let api: OpenAPI.Document = await oas.validate({ convertToLatest: opts.convertToLatest }).catch((err: Error) => {
+  let api: OpenAPI.Document;
+  await oas.validate().catch((err: Error) => {
     spinner.fail();
     debug(`raw validation error object: ${JSON.stringify(err)}`);
     throw err;
   });
+
+  // If we were supplied a Postman collection this will **always** convert it to OpenAPI 3.0.
```
kanadgupta, line 193: mind adding a debug statement here, sort of like how we're doing on line 202?

kanadgupta (approved):
a few non-blocking comments below otherwise LGTM!

src/lib/prepareOas.ts
erunion: We don't, but this thing method seems to always want to return an OpenAPI definition as it's mutating `info.title`.

kanadgupta (approved):
this got me thinking about a few issues we should figure out soon, see https://github.com/readmeio/rdme/pull/1141 and https://github.com/readmeio/rdme/issues/1142

PR #1127, opened 2024-12-21
feat(openapi): a fresh coat of paint on `openapi inspect`

erunion:
## 🧰 Changes

The biggest problem I've had since originally building our `openapi inspect` command has been that the tables between OpenAPI and ReadMe features aren't of consistent widths. Drives me up a damn wall.

Screenshot 2024-12-20 at 9 50 05 PM

So I'm giving a fresh coat of paint to this command that deserves some love.

* [x] Consistent table widths. <sub>existential grumbling</sub>
* [x] Table headers are now yellow instead of green to reduce the amount of color thrashing we have going on.
* [x] URLs to OpenAPI spec definitions our own docs are now dimmed instead of grey.
* [x] Tossed an emoji on the different feature sections. 🌲 for OpenAPI and 📖 for us.
* [x] Swapped out the ✅ emoji for used features in place of 🟢 and 🔴 circles.
* [x] Usage strings like "2 Media Types", "a total of 9 Operations", or "a single Security Type" are now lowercased because it looked funky as hell.
* [x] Bumped OpenAPI spec links to the latest versions for 3.0 and 3.1.

I also made a couple minor readability to how we construct this table because the code was a bit gnar with some deeply nested `array.push()` calls.

## 🧬 QA & Testing

This is what it all looks like:

Screenshot 2024-12-20 at 9 42 57 PM


src/commands/openapi/inspect.ts
```diff
@@ -108,68 +108,71 @@ function buildFullReport(analysis: Analysis, definitionVersion: string, tableBor
...
+            const pluralName = pluralize(info.name.toLowerCase(), info.found.length);
+            return `You are using ${chalk.bold(info.found.length)} ${pluralName} throughout your API: ${new Intl.ListFormat('en').format(highlightedData)}`;
+          }
+
+          return `You are using a single ${info.name.toLowerCase()} throughout your API: ${highlightedData[0]}`;
+        } else if (info.found > 1) {
+          let msg: string;
+          msg = `You have a total of ${chalk.bold(info.found)} ${pluralize(info.name.toLowerCase(), info.found)} in your API.`;
+          if (info.found > 100) {
+            msg += ` ${chalk.cyanBright('Wow! 🤯')}`;
```
erunion, line 128: iffy on the emoji but thought it might be fun for the true Big API sickos if this stood out a bit more. also might bump this 100 operation threshold up to maybe 200 because after this week of debugging failed API uploads 100 operations in a single file seems so small

Screenshot 2024-12-20 at 9 44 03 PM

src/commands/openapi/inspect.ts
```diff
@@ -108,68 +108,71 @@ function buildFullReport(analysis: Analysis, definitionVersion: string, tableBor
...
-
-    report.push('');
-    report.push(header);
+    report.push(`${emoji} ${header}`);
     report.push(
       table(tableData, {
         border: tableBorder,
         columns: {
+          0: {
+            width: 26,
```
erunion, line 174: This caps the "Feature" table column to a specific width, which because the second "Used" one doesn't have variable content, causes the width of every table to now be identical.

src/commands/openapi/inspect.ts
```diff
@@ -108,68 +108,71 @@ function buildFullReport(analysis: Analysis, definitionVersion: string, tableBor
...
-  });
+  report.push(''); // Extra bit of space between our general API information and feature tables.
 
   // Build out a view of all OpenAPI and ReadMe features that we discovered.
   [
-    { component: 'openapi', header: 'OpenAPI Features' },
-    { component: 'readme', header: 'ReadMe-Specific Features and Extensions' },
-  ].forEach(({ component, header }: { component: string; header: string }) => {
+    { component: 'openapi', header: 'OpenAPI Features', emoji: '🌲' },
+    { component: 'readme', header: 'ReadMe-Specific Features and Extensions', emoji: '📖' },
```
kanadgupta, line 145: don't feel strongly here but this feels a little bit more unambiguously readme:

```suggestion
    { component: 'readme', header: 'ReadMe-Specific Features and Extensions', emoji: '🦉' },
```

src/commands/openapi/inspect.ts
kanadgupta: obsessed with this

kanadgupta (approved):

src/commands/openapi/inspect.ts
erunion: ended up bumping this up to 200

domharrington (approved):

kanadgupta (approved):

PR #1128, opened 2024-12-21
chore: cleaning up some type imports

erunion:
## 🧰 Changes

This just cleans up some `import type { typeName} from ...` imports so that they're imported that way instead of as `import { type typeName }` from ...`. We generally use `import type` everywhere already there's just some stragglers that sometimes get through the other way via VSCode.


kanadgupta (changes requested):
my vscode auto-import setup (which i believe is fairly standard) did this so i'm sure this will pop up again 😮‍💨 tbh, without lint rules that enforce these stylistic choices, i don't think it makes sense to maintain them

PR #1141, opened 2025-01-06
fix(openapi): type fixes in `prepareOas`

kanadgupta:
## 🧰 Changes

when reviewing https://github.com/readmeio/rdme/pull/1126, i noticed a few small bugs in our `prepareOas` helper function. this PR makes a few typesafety improvements so we shouldn't run into this issue again(?)

~~one outstanding question that i'd like to figure out prior to merge, see https://github.com/readmeio/rdme/issues/1142~~ (**edit**: going to continue bundling and handle this issue in a follow-up PR)

## 🧬 QA & Testing

do tests + types pass?

### DIFF
--- a/src/lib/prepareOas.ts
+++ b/src/lib/prepareOas.ts
@@ -36,6 +36,8 @@ function truthy<T>(value: T): value is Truthy<T> {
   return !!value;
 }
 
+type OpenAPIAction = 'convert' | 'inspect' | 'reduce' | 'upload' | 'validate';
+
 const capitalizeSpecType = (type: string) =>
   type === 'openapi' ? 'OpenAPI' : type.charAt(0).toUpperCase() + type.slice(1);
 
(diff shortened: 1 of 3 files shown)


src/lib/prepareOas.ts
```diff
@@ -213,7 +209,9 @@ export default async function prepareOas(
   const specVersion: string = api.info.version;
   debug(`version in spec: ${specVersion}`);
 
-  if (['openapi', 'openapi inspect', 'openapi reduce'].includes(command)) {
+  const commandsThatBundle: (typeof command)[] = ['openapi inspect', 'openapi reduce', 'openapi upload'];
```
kanadgupta, line 212: 🧐 do we want to bundle by default for `openapi upload`? see https://github.com/readmeio/rdme/issues/1142

erunion: i think we have to because if we don't then they might upload an invalid spec that has ref pointers to a local file

erunion (approved):

PR #1153, opened 2025-01-25
fix(openapi): adding support for OpenAPI 3.1.1, 3.0.4 and basic auth fetches

### LINKED ISSUE #1152: CLI doesn't support basic authentication in the URL since 9.0.0
Since this PR: https://github.com/readmeio/rdme/pull/915, where the CLI moved to use `oas-normalize.load`, basic authentication stopped working.

That is because in oas-normalize, we use `fetch` directly on the URL: https://github.com/readmeio/oas/blob/main/packages/oas-normalize/src/index.ts#L66 , which doesn't support basic authentication in the URL.

erunion:
| 🚥 Resolves https://github.com/readmeio/rdme/issues/1152 |
| :------------------- |

## 🧰 Changes

This upgrades `oas-normalize` and `oas` in order to pull in updates made to our OpenAPI parser to support OpenAPI 3.0.4 and 3.1.1 definitions. I have also made a fix in `oas-normalize` to support basic auth credentials in URLs, which will fix #1152.

Also upgraded `nock` to v14 while I'm here because that dropped today.


package.json
```diff
@@ -72,7 +72,7 @@
     "@commitlint/cli": "^19.0.3",
     "@commitlint/config-conventional": "^19.0.3",
     "@oclif/test": "^4.1.0",
-    "@readme/better-ajv-errors": "^1.5.0",
+    "@readme/better-ajv-errors": "^2.0.0",
```
erunion, line 75: This got a major bump because I dropped support for Node 16 (finally).

kanadgupta (approved):

Issue #1142, opened 2025-01-06
Confusing uploading behavior for YAML files


kanadgupta:
In `openapi upload`, we currently always bundle (with the fix in https://github.com/readmeio/rdme/pull/1141, at least) the OAS and convert it to JSON prior to sending it to ReadMe.

This creates unexpected behavior for YAML files, where it's possible to send a YAML file and we'll send the file with a `.yml` file extension, yet the file blob itself is JSON due to our converting/bundling.

How should we handle this? Should we ever send the original, unbundled, unconverted file?

emilyskuo:
Looked into this and confirmed that YAML files are being uploaded to ReadMe in YAML. The OAS gets converted to JSON when viewed in ReadMe.

PR #1154, opened 2025-01-25
fix(openapi/upload): improved handling for YAML files

emilyskuo:
| 🚥 Resolves RM-11893, https://github.com/readmeio/rdme/issues/1142 |
| :------------------- |

## 🧰 Changes

When uploading an API definition, `rdme` converts all files to JSON, which means that YAML files are uploaded in JSON but with the file extension of `.yaml` or `.yml`. This will convert files originally in YAML back to YAML before uploading to ReadMe.

## 🧬 QA & Testing

Upload a spec in YAML and confirm in GitHug that it's been stored in YAML. Also confirm that JSON specs are uploaded in JSON as well.


emilyskuo:
Closing this out since YAML files are being uploaded in YAML, and the conversion to JSON is happening on the ReadMe side when the spec is viewed.

emilyskuo:
Found out that I was testing an edge case, and that YAML files are not always uploaded in YAML. Reopening this PR to fix that.

package.json
```diff
@@ -68,6 +68,7 @@
     "table": "^6.8.1",
     "tmp-promise": "^3.0.3",
     "toposort": "^2.0.2",
+    "@types/js-yaml": "^4.0.5",
```
erunion, line 71: can this get moved back to being a devdep?

erunion (approved):

package.json
emilyskuo: I got a lint error when I left it in devdeps saying it needed to be in the regular dependencies
Screenshot 2025-02-13 at 3 30 10 PM

emilyskuo: When I move it to devdeps and run `npm run build`, then `npm run lint`, I still get the same error 😰

emilyskuo:
@kanadgupta I couldn't figure out a good way to test the format of the upload itself since that's not an output of the command. Instead I updated one of the logs to include the file type - lmk if that's okay?

package.json
kanadgupta: @emilyskuo looked into this and i think i figured out the issue — the linting error is a bit deceptive. `js-yaml` should be a regular dep and `@types/js-yaml` should remain a dev-dep. mind making that change?

kanadgupta (changes requested):

package.json
emilyskuo: Ah that makes more sense. Updated, thanks!

__tests__/commands/openapi/upload.test.ts
```diff
@@ -50,6 +52,26 @@ describe('rdme openapi upload', () => {
...
+        )
+        .reply(200, {
+          data: {
+            upload: { status: 'done' },
+            uri: `/versions/${version}/apis/${slugifiedYamlFile}`,
+          },
+        });
+
+      const result = await run(['--version', version, yamlFile, '--key', key]);
+      expect(result.stdout).toContain('was successfully created in ReadMe!');
```
kanadgupta, line 70: mind making this a snapshot assertion?

```suggestion
      expect(result).toMatchSnapshot();
```

package.json
```diff
@@ -56,6 +56,7 @@
     "debug": "^4.3.3",
     "gray-matter": "^4.0.1",
     "ignore": "^7.0.0",
+    "js-yaml": "^4.1.0",
```
kanadgupta, line 59: i _think_ this change should result in a change to our lockfile. mind running `npm i` and committing the changes?

kanadgupta (commented):
two more small changes, but otherwise LGTM!

erunion (approved):

__tests__/commands/openapi/__snapshots__/upload.test.ts.snap
kanadgupta, line 1: i _think_ this should have updated a snapshot rather than creating a new one — if you run this test file, does vitest say that there are any obsolete snapshots that can be removed?

kanadgupta (approved):

__tests__/commands/openapi/__snapshots__/upload.test.ts.snap
emilyskuo: Weird, yeah there was an obsolete snapshot. Removed!

kanadgupta (approved):
🚀

PR #1165, opened 2025-02-03
chore(vscode): configure search exclusions

kanadgupta:
## 🧰 Changes

when pairing with @emilyskuo we noticed a handful of directories/files that i had configured myself but didn't configure for the repo writ large. figured all vscode users could benefit from this config.

## 🧬 QA & Testing

do the search results properly exclude certain directories/files?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
   "search.exclude": {
     "**/dist-gha": true,
     "CHANGELOG.md": true,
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "search.exclude": {
```
erunion, line 2: want to add these too?

```suggestion
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "editor.codeActionsOnSave": {
    "source.fixAll": "explicit"
  },
  "editor.formatOnSave": true,
  "search.exclude": {
```

erunion (approved):

.vscode/settings.json
kanadgupta: i personally don't like those first two but i'll configure prettier yeah

Issue #1167, opened 2025-02-06
openapi upload doesn't support skipping overwrite prompt


jarrodparkes:
I've been unable to find a way to skip the following prompt:

```bash
npx rdme@10 openapi upload PATH_TO_SPEC --key=KEY --slug=SLUG --useSpecVersion
✔ This will overwrite the existing API definition for Aryeo.json. Are you sure you want to continue?
```

I could understand if this warning was unavoidable IF I could alternatively use the CLI to delete an existing API definition, but that also seems impossible.

Is this (1) a missing feature or (2) a case of improper use?

kanadgupta:
Agreed that this would be useful to have — appreciate you taking the time to open up

PR #1168, opened 2025-02-07
feat(openapi/upload): add `--confirm-overwrite` flag

jarrodparkes:
| 🚥 Resolves https://github.com/readmeio/rdme/issues/1167 |
| :------------------- |

## 🧰 Changes

If an automated system runs `openapi upload` on a pre-existing file, then an overwrite prompt is shown. There is no way to skip this prompt which can lock up automated systems.

This PR adds a flag to skip the prompt.

## 🧬 QA & Testing

**Case 1 - confirm existing behavior**
- Upload a new API spec to readme using `openapi upload {spec} --key={key} --slug={slug}`
- Try uploading the API spec again to trigger the overwrite prompt

**Case 2 - test new behavior**
- Upload a new API spec to readme using `openapi upload {spec} --key={key} --slug={slug}`
- Try uploading the API spec again using `openapi upload {spec} --key={key} --slug={slug} --overwrite` to skip the overwrite prompt

### DIFF
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -29,9 +29,9 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
 
   static flags = {
     key: keyFlag,
-    overwrite: Flags.boolean({
+    'confirm-overwrite': Flags.boolean({
       description:
-        'If set, the overwrite prompt will be skipped. This flag can be a useful in automated environments where prompts cannot be responded to.',
+        'If set, file overwrites will be made without a confirmation prompt. This flag can be a useful in automated environments where prompts cannot be responded to.',
       hidden: true,
     }),
     slug: Flags.string({
@@ -142,7 +142,7 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
     if (method === 'PUT') {
       // bypass the prompt if we're in a CI environment
       prompts.override({
-        confirm: isCI() ? true : this.flags.overwrite,
+        confirm: isCI() || this.flags['confirm-overwrite'] ? true : undefined,
       });
 
       const { confirm } = await promptTerminal({
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -29,6 +29,11 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
 
   static flags = {
     key: keyFlag,
+    overwrite: Flags.boolean({
+      description:
+        'If set, the overwrite prompt will be skipped. This flag can be a useful in automated environments where prompts cannot be responded to.',
+      hidden: true,
+    }),
     slug: Flags.string({
       summary: 'Override the slug (i.e., the unique identifier) for your API definition.',
       description: [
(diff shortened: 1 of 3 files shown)


src/commands/openapi/upload.ts
```diff
@@ -29,6 +29,11 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
 
   static flags = {
     key: keyFlag,
+    overwrite: Flags.boolean({
```
kanadgupta, line 32: Can you rename this flag?

```suggestion
    'confirm-overwrite': Flags.boolean({
```

src/commands/openapi/upload.ts
```diff
@@ -137,7 +142,7 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
     if (method === 'PUT') {
       // bypass the prompt if we're in a CI environment
       prompts.override({
-        confirm: isCI() ? true : undefined,
+        confirm: isCI() ? true : this.flags.overwrite,
```
kanadgupta, line 145: I _think_ this is might be a bug but I might be wrong. I believe if `this.flags.overwrite` is `false`, the value of `confirm` is overridden to `false`, which skips the prompt and throws the error on line 155. Can you confirm? I believe this is how we'd fix it:

```suggestion
        confirm: isCI() || this.flags.overwrite ? true : undefined,
```

kanadgupta (changes requested):
Appreciate the contribution, @jarrodparkes! Can you add a test? You can find all of our tests in `__tests__/commands/openapi/upload.test.ts`. Let us know if you need a hand. Thanks!

src/commands/openapi/upload.ts
jarrodparkes: @kanadgupta 

I made the recommended change (I had to modify the arg name), and it worked as desired — 

1. when the flag is provided, the prompt is skipped
2. when the flag is omitted, the prompt is shown

jarrodparkes:
whoops, forgot to add a test. i'll do that momentarily!

kanadgupta:
@jarrodparkes please re-request me once you've added a test!

jarrodparkes:
@kanadgupta I cannot figure out how to confirm the test i've written. i've tried a few different ways to start the test suite, or to test the individual test file, but i'm not having any luck.

i see output like this:

```
> rdme@10.2.0-next.1 test
> vitest run --coverage


 RUN  v3.0.5 /Users/jarrodp/Dev/rdme
      Coverage enabled with v8


 Test Files 0 passed (20)
      Tests 0 passed (0)
   Start at 10:48:00
   Duration 103ms
```

which seems to suggest to me that nothing is actually running?

kanadgupta:
A few questions:

1. What version of Node.js are you running? We require at least Node v20 or higher:

https://github.com/readmeio/rdme/blob/76f488014e1f179c405a11b8f1bd8234185f9e45/package.json#L7-L9

2. Once you've verified that you're using the correct Node.js version, try running the test file using the following commands:

```sh
# install + build
npm ci
npm run build

# run the test file
npx vitest __tests__/commands/openapi/upload.test.ts
```

kanadgupta (approved):
Thanks @jarrodparkes!

PR #1172, opened 2025-02-19
ci: unify GHA build process

kanadgupta:
## 🧰 Changes

a few small behind-the-scenes changes to simplify how we build `rdme` for GitHub Actions users — no actual customer impact here. quick breakdown of the changes:

- [x] replaces `@rollup/plugin-terser` with `rollup-plugin-esbuild` for minifying our GitHub Actions code. the latter accomplishes basically the same result as the former, but in a tiny fraction of the time.
- [x] we previously forked our GHA build process into a development vs. production setup since terser takes so long. since `rollup-plugin-esbuild` adds a negligible amount of time to our build process, there's no need to have two setups. this PR combines them into a single build process.
- [x] added a bunch of steps to our CI to further assert that the GitHub Actions build is what we expect.

## 🧬 QA & Testing

does CI still pass?


.github/workflows/ci.yml
```diff
@@ -98,11 +103,28 @@ jobs:
...
+      # This step should fail because `help` does not exist in GitHub Actions
+      - name: Run `help`
+        uses: ./rdme-repo/
+        id: help-fail
+        continue-on-error: true
+        with:
+          rdme: help
+
+      - name: Assert that previous step failed
+        if: ${{ steps.help-fail.outcome == 'failure' }}
```
erunion, line 122: neat

rollup.config.js
```diff
@@ -1,11 +1,10 @@
+// @ts-check
```
erunion, line 1: can this file be TS?

erunion (approved):

rollup.config.js
kanadgupta: it requires installing a whole add'l dependency so i opted not to: https://rollupjs.org/command-line-interface/#config-intellisense

PR #1174, opened 2025-02-21
chore(deps-dev): upgrading our eslint standards

erunion:
## 🧰 Changes

The owner of `eslint-plugin-vitest` lost their NPM access to it was moved into `@vitest/eslint-plugin` a little over a year ago. This upgrades our `@readme/eslint-config` package to pull that in.

### DIFF
--- a/__tests__/.eslintrc
+++ b/__tests__/.eslintrc
@@ -6,7 +6,5 @@
     "unicorn/prefer-module": "off",
 
     "@vitest/require-mock-type-parameters": "off",
-    "@vitest/padding-around-expect-groups": "off",
-    "@vitest/padding-around-all": "off",
   }
 }
--- a/__tests__/.eslintrc
+++ b/__tests__/.eslintrc
@@ -3,6 +3,10 @@
   "rules": {
     // vitest currently allows for __dirname, require, etc.
     // see https://github.com/vitest-dev/vitest/issues/2841
-    "unicorn/prefer-module": "off"
+    "unicorn/prefer-module": "off",
+
+    "@vitest/require-mock-type-parameters": "off",
+    "@vitest/padding-around-expect-groups": "off",
+    "@vitest/padding-around-all": "off",
   }
 }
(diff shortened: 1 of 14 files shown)


__tests__/.eslintrc
```diff
@@ -3,6 +3,10 @@
   "rules": {
     // vitest currently allows for __dirname, require, etc.
     // see https://github.com/vitest-dev/vitest/issues/2841
-    "unicorn/prefer-module": "off"
+    "unicorn/prefer-module": "off",
+
+    "@vitest/require-mock-type-parameters": "off",
+    "@vitest/padding-around-expect-groups": "off",
+    "@vitest/padding-around-all": "off",
```
kanadgupta, line 10: why are these disabled?

__tests__/lib/fetch.test.ts
kanadgupta, line 1: what are the changes in this file for? mind reverting? we use this `mock.done()` pattern pretty widely in this codebase and i see no issues with it

__tests__/lib/createGHA.test.ts
```diff
@@ -123,7 +123,7 @@ describe('#createGHA', () => {
         });
 
         const res = await oclifConfig.runHook('createGHA', { command: CurrentCommand, parsedOpts: opts, result: '' });
-        expect(res.successes[0].result).toBeTruthy();
+        expect(res.successes[0].result).toBe(true);
```
kanadgupta, line 126: did eslint catch this?

__tests__/.eslintrc
erunion: disabled the first one because typing `vi.fn()` was becoming a huge pain.

disabled the last two because it wants to update a whole bunch of tests to add padding around `expect()` calls. can enable them if you like

__tests__/lib/fetch.test.ts
erunion: made these changes because this test was being flagged as not having any expectations so i rewrote it to use `mock.isDone()` as `mock.done()` throws an exception if a nock mock wasn't called. happy to revert and ignore the rule here

__tests__/lib/createGHA.test.ts
erunion: yea

kanadgupta (approved):

__tests__/.eslintrc
erunion: nah im always pro whitespace. will enable it

kanadgupta (approved):

PR #1175, opened 2025-02-25
fix(docs/upload): add safeguard if projects have bidi enabled

emilyskuo:
| 🚥 Resolves RM-11901 |
| :------------------- |

## 🧰 Changes

Adding in a safe guard for ReadMe projects with bidirectional sync enabled, since the docs frontmatter format is different for bidi. By default, `rdme docs upload` may try to "fix" things and modify the markdown files, which could result in corrupted data and/or syncing errors.

Once https://github.com/readmeio/readme/pull/13869 is merged, we'll be exposing info on the project's bidi connection if it exists via the get project metadata (`/projects/me`) endpoint. This PR will introduce a check to see if a bidi connection exists, and throw an error if it does. If users want to continue uploading via `rdme`, they can use the `--skip-validation` flag to do so without potential issues from `rdme`'s validation & fixing steps. 

## 🧬 QA & Testing

Tested this against https://github.com/readmeio/readme/pull/13869 locally. Added some tests for `docs upload` as well.


src/lib/types.ts
kanadgupta, line 1: 😮‍💨 do you think we could make it so we split out the openapi changes into their own PR? and this PR is branched off of that PR?

src/lib/syncPagePath.ts
```diff
@@ -208,6 +208,23 @@ export default async function syncPagePath(this: CommandsThatSyncMarkdown) {
...
+    method: 'GET',
+    headers,
+  }).then(res => {
+    return this.handleAPIRes(res);
+  });
+
+  const biDiConnection = projectMetadata.data.git?.connection;
+  if (biDiConnection && !skipValidation) {
+    throw new Error(
+      "Bi-directional syncing is enabled for this project. Uploading these docs will overwrite what's currently synced from Git. To proceed with uploading via `rdme`, please use the `--skip-validation` flag.",
```
kanadgupta, line 224:
```suggestion
      "Bi-directional syncing is enabled for this project. Uploading these docs will overwrite what's currently synced from Git. To proceed with the upload, re-run this command with the `--skip-validation` flag.",
```

src/lib/syncPagePath.ts
```diff
@@ -208,6 +208,23 @@ export default async function syncPagePath(this: CommandsThatSyncMarkdown) {
...
+    );
+  }
+
   if (skipValidation) {
```
kanadgupta, line 228: thoughts on making it so this warning is only shown if bidi is **not** enabled? or maybe we have a separate warning for those users?

```suggestion
  if (skipValidation && !bidiConnection) {
```

src/lib/syncPagePath.ts
```diff
@@ -208,6 +208,23 @@ export default async function syncPagePath(this: CommandsThatSyncMarkdown) {
...
+  // if it is, validations must be skipped to prevent frontmatter from being overwritten
+  const headers = new Headers({ authorization: `Bearer ${key}` });
+  const projectMetadata: ProjectRepresentation = await this.readmeAPIFetch('/projects/me', {
+    method: 'GET',
+    headers,
+  }).then(res => {
+    return this.handleAPIRes(res);
+  });
+
+  const biDiConnection = projectMetadata.data.git?.connection;
```
kanadgupta, line 221: maybe just to be extra cautious:

```suggestion
  const biDiConnection = projectMetadata?.data?.git?.connection;
```

__tests__/commands/docs/upload.test.ts
```diff
@@ -24,6 +24,9 @@ describe('rdme docs upload', () => {
       categories: {},
       parentPages: {},
     });
+    getAPIv2Mock({ authorization }).get('/projects/me').reply(200, {
```
kanadgupta, line 27: these tests are great, ty! could you also add a test case for if this initial `/projects/me` request fails? i'll DM you an example error response that you can pull from shortly

kanadgupta (changes requested):
a few small outstanding things but this looks great on the whole!

src/lib/syncPagePath.ts
emilyskuo: If we had a separate warning for `skipValidation` + bidi, what would we wanna say? Should we repeat the warning from the bidi without `skipValidation` error? (ie "Bi-directional syncing is enabled for this project. Uploading these docs will overwrite what's currently synced from Git.")

src/lib/types/openapiDoc.ts
```diff
@@ -3245,6 +3245,21 @@ const document = {
                           description:
                             'The description of your project. This is used in the page meta description and is seen by search engines and sites like Facebook.',
                         },
+                        git: {
```
emilyskuo, line 3248: Temporarily adding this in manually instead of generating the doc again - will do that once https://github.com/readmeio/readme/pull/13869 is good and merged

src/lib/syncPagePath.ts
kanadgupta: yep i think that's perfect!

src/lib/types.ts
emilyskuo: New branch here for the openapi changes! https://github.com/readmeio/rdme/pull/1195

src/lib/types/openapiDoc.ts
emilyskuo: See https://github.com/readmeio/rdme/pull/1195

kanadgupta (approved):

PR #1212, opened 2025-04-14
chore(deps): upgrading openapi-related tooling

erunion:
## 🧰 Changes

This upgrades our OpenAPI tooling in order to pull in a fix for a serious performance issue we uncovered when dereferencing API definitions that make extensive use of circular references. Check out https://github.com/APIDevTools/json-schema-ref-parser/pull/380 for details.


emilyskuo (approved):

__tests__/commands/openapi/__snapshots__/inspect.test.ts.snap
```diff
@@ -93,13 +103,8 @@ additionalProperties:
  · #/components/schemas/ParameterizedHeader/properties/parameters/additionalProperties
 
 circularRefs:
- · #/components/schemas/BodyPart/properties/parent
```
erunion, line 96: These snapshots got updated because of the circular reference performance change in `@apidevtools/json-schema-ref-parser`.

PR #1213, opened 2025-04-16
feat(cli): add support for plugins

kanadgupta:
## 🧰 Changes

adds initial support for `rdme` plugins. not going to write any further docs on this for the time being. more context can be found [here](https://readmeio.slack.com/archives/C08NTUB6XPB/p1744826658541339).

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@oclif/plugin-autocomplete": "^3.2.6",
     "@oclif/plugin-help": "^6.2.15",
     "@oclif/plugin-not-found": "^3.2.28",
+    "@oclif/plugin-plugins": "^5.4.36",
     "@oclif/plugin-warn-if-update-available": "^3.1.19",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -170,8 +171,11 @@
       "@oclif/plugin-autocomplete",
       "@oclif/plugin-help",
       "@oclif/plugin-not-found",
+      "@oclif/plugin-plugins",
       "@oclif/plugin-warn-if-update-available"
     ],
+    "pluginPrefix": "rdme-plugin",
+    "scope": "readme",
```
kanadgupta, line 178: more info here: https://github.com/oclif/plugin-plugins?tab=readme-ov-file#friendly-names

also i opened up a PR since the docs don't mention `pluginPrefix` but it's definitely a thing: https://github.com/oclif/plugin-plugins/pull/1113

PR #1216, opened 2025-04-18
feat: various changes to get us plugin-ready

kanadgupta:
## 🧰 Changes

this PR handles a lot of housekeeping to lay the groundwork for the upcoming internal plugins project.

the biggest change is that we're now building declaration files and shipping types for consumption by other plugins. type imports will look something like this:

```ts
import type { PageMetadata, PluginHooks } from 'rdme/types';
```

beyond that, it's mostly little bug fixes, formatting tweaks, and test coverage enhancements. i also removed `@oclif/plugin-not-found` since the plugin makes it brutally slow if you typo something — figured we can add it back if folks complain but i'd be surprised if that was the case.

## 🧬 QA & Testing

do tests/etc. pass?


__tests__/commands/docs/__snapshots__/upload.test.ts.snap
```diff
@@ -206,6 +206,43 @@ exports[`rdme docs upload > given that the file path is a directory > should han
 }
 `;
 
+exports[`rdme docs upload > given that the file path is a directory > should handle complex frontmatter 1`] = `
```
davinhazard, line 209:
```suggestion
exports[`rdme docs upload > given that the file path is a directory > should handle complex frontmatter`] = `
```

kanadgupta: ah, this is an autogenerated [vitest snapshot](https://vitest.dev/guide/snapshot.html#use-snapshots)! when i attempt to change it, vitest switches it back

davinhazard (approved):
✨

__tests__/commands/docs/__snapshots__/upload.test.ts.snap
davinhazard: oh right weird

package.json
```diff
@@ -45,7 +48,6 @@
     "@oclif/core": "^4.0.29",
     "@oclif/plugin-autocomplete": "^3.2.6",
     "@oclif/plugin-help": "^6.2.15",
-    "@oclif/plugin-not-found": "^3.2.28",
```
erunion, line 48: how come we dont need this anymore?

src/lib/readmeAPIFetch.ts
```diff
@@ -475,7 +479,7 @@ export async function fetchMappings(this: CommandClass['prototype']): Promise<Ma
 /**
  * Fetches the schema for the current route from the OpenAPI description for ReadMe API v2.
  */
-export async function fetchSchema(this: CommandsThatSyncMarkdown): Promise<SchemaObject> {
+export async function fetchSchema(this: CommandsThatSyncMarkdown) {
```
erunion, line 482: we not running with isolatedDeclarations?

erunion (approved):

package.json
kanadgupta: per the PR description:

> i also removed `@oclif/plugin-not-found` since the plugin makes it brutally slow if you typo something — figured we can add it back if folks complain but i'd be surprised if that was the case.

src/lib/readmeAPIFetch.ts
kanadgupta: ah we probably should now, i'll enable that

kanadgupta: ugh it's a lot more work to enable it in this repo than i have time for atm. will look into it in a follow-up

PR #1218, opened 2025-04-21
fix: fetch util types

kanadgupta:
## 🧰 Changes

widens the types of these util functions a bit so they play more nicely with plugins.

## 🧬 QA & Testing

i confirmed locally that these types play nicely with plugins.

### DIFF
--- a/src/lib/readmeAPIFetch.ts
+++ b/src/lib/readmeAPIFetch.ts
@@ -228,6 +228,10 @@ export async function readmeAPIv1Fetch(
  * Wrapper for the `fetch` API so we can add rdme-specific headers to all ReadMe API v2 requests.
  */
 export async function readmeAPIv2Fetch<T extends Hook.Context = Hook.Context>(
+  /**
+   * `this` does not have to be a hook, it can also be a Command class.
+   * This type ensures that `this` has the `config` and `debug` properties.
+   */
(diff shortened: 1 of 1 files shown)


src/lib/readmeAPIFetch.ts
```diff
@@ -367,8 +367,8 @@ export async function handleAPIv1Res(
  *
  * If we receive non-JSON responses, we consider them errors and throw them.
  */
-export async function handleAPIv2Res(
-  this: CommandClass['prototype'] | Hook.Context,
+export async function handleAPIv2Res<T extends Hook.Context = Hook.Context>(
```
erunion, line 370: `Hook.Context` also works for `CommandClass['prototype']`?

erunion (approved):

src/lib/readmeAPIFetch.ts
kanadgupta: it's actually quite small and generic and contains just `config` and logging properties like a command does:

https://github.com/oclif/core/blob/d4728f764350907d1187f837e8190f5b96285536/src/interfaces/hooks.ts#L129

kanadgupta: i'll add a little comment so that's clearer

PR #1223, opened 2025-04-22
chore: updating our OpenAPI schema

erunion:
## 🧰 Changes

Updates our OpenAPI schema with some minor revisions to our public API.

### DIFF
--- a/src/lib/types/openapiDoc.ts
+++ b/src/lib/types/openapiDoc.ts
@@ -15,7 +15,7 @@ const document = {
     version: '2.0.0-beta',
     title: 'ReadMe API v2 🦉 (BETA)',
     // @ts-expect-error custom extension
-    'x-readme-deploy': '5.334.1',
+    'x-readme-deploy': '5.336.0',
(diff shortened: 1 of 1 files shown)


src/lib/types/openapiDoc.ts
```diff
@@ -6469,36 +6593,6 @@ const document = {
                             nullable: true,
                             description: 'A non-semver display name for the version.',
                           },
-                          git: {
```
kanadgupta, line 6472: why was this removed?

kanadgupta (approved):

src/lib/types/openapiDoc.ts
erunion: @julshotal Refactored it away as part of the branching project

julshotal: we don't use it on the front end & we figured when it comes to merging functionality later saving the `latestCommitDate` here wasn't something we felt confident about always being the most updated data

PR #1225, opened 2025-04-22
feat: exposing types for creating API keys to plugins

erunion:
## 🧰 Changes

Exposes a new `APIKeyRepresentation` type for use in plugins.


src/types.ts
```diff
@@ -1,4 +1,4 @@
-export type { GuidesRequestRepresentation, ProjectRepresentation } from './lib/types/index.js';
+export type { APIKeyRepresentation, GuidesRequestRepresentation, ProjectRepresentation } from './lib/types/index.js';
```
erunion, line 1: Beginning to wonder if we want to rework how we're doing these and just make **everything** available rather than piecemeal exposings.

kanadgupta: i personally prefer piecemeal so as to make sure we're prescriptive about what is fair game to use in plugins, but open to changing my mind about that

kanadgupta (approved):

PR #1227, opened 2025-04-22
fix: timeouts on `doc:upload`

erunion:
## 🧰 Changes

We've been running into issues with `doc:upload` calls processing 6 pages and then failing out with undici timeout errors despite no actual request ever happening. I don't really know what's going on with the `allSettled` or undici and why it's consistently failing but reworking `doc:upload` to process each page in the batch one by one resolves our issues (albeit at the expensive of making the command slower).

### DIFF
--- a/src/lib/readmeAPIFetch.ts
+++ b/src/lib/readmeAPIFetch.ts
@@ -322,6 +322,7 @@ export async function readmeAPIv2Fetch<T extends Hook.Context = Hook.Context>(
     })
     .catch(e => {
       this.debug(`error making fetch request: ${e}`);
+      this.debug(e.stack);
       throw e;
     });
 }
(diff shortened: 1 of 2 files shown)


src/lib/readmeAPIFetch.ts
```diff
@@ -322,6 +322,7 @@ export async function readmeAPIv2Fetch<T extends Hook.Context = Hook.Context>(
     })
     .catch(e => {
       this.debug(`error making fetch request: ${e}`);
+      this.debug(e.stack);
```
erunion, line 325: Added this in because it was annoying only setting `fetch failed` errors with no additional context.

kanadgupta (approved):
LGTM sans the lint error

erunion:
Not having VSCode set up in this repo to autofix errors on file save gets me every damn time.

PR #1228, opened 2025-04-23
chore: updating our OpenAPI schema

erunion:
## 🧰 Changes

Updates our OpenAPI schema as we've temporarily turned off our `GET /versions` endpoint while we rework it to support some in-flight projects that are reworking int.


kanadgupta (commented):
this technically hasn't hit prod yet — cool if we hold off on merging this until the `x-readme-deploy` value reflects this change?

kanadgupta (approved):

PR #1230, opened 2025-04-23
feat(docs/migrate): migration stats

kanadgupta:
| 🚥 Resolves RM-12506 |
| :------------------- |

## 🧰 Changes

adds migration stats reporting to the `docs:migrate` command. this was a big exercise in strong types to ensure that plugin authors have enough guardrails when writing these stats for themselves.

as part of this:
- [x] added a bunch of types and a tiny bit of logic (which is technically a breaking change) to the `pre_markdown_validation` hook return types so it includes both the transformed pages as well as the stats results
- [x] added additional type safety to our syncing results (so we can use these more effectively in downstream plugins)
- [x] added a hidden `--max-errors` flag (inspired in part by [eslint's `--max-warnings` flag](https://eslint.org/docs/latest/use/command-line-interface#--max-warnings)) to `docs:upload` so users/plugin authors have the ability to control how errors are returned

## 🧬 QA & Testing

do tests + types pass?

### DIFF
--- a/src/commands/docs/migrate.ts
+++ b/src/commands/docs/migrate.ts
@@ -1,4 +1,4 @@
-import type { PluginHooks } from '../../lib/hooks/exported.js';
+import type { MigrationStats, PluginHooks } from '../../lib/hooks/exported.js';
 
 import { Args, Flags, type Hook } from '@oclif/core';
 import chalk from 'chalk';
@@ -46,7 +46,7 @@ export default class DocsMigrateCommand extends BaseCommand<typeof DocsMigrateCo
     }),
   };
 
-  async run() {
+  async run(): Promise<{ outputDir: string; stats: MigrationStats }> {
     if (!this.flags['hide-experimental-warning']) {
       this.warn(alphaNotice);
     }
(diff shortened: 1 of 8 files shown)


src/commands/docs/migrate.ts
```diff
@@ -76,6 +76,16 @@ export default class DocsMigrateCommand extends BaseCommand<typeof DocsMigrateCo
       }
     });
 
+    const stats: MigrationStats = {
+      migrationOutputDir: outputDir,
+      results: {},
+      timestamp: new Date().toISOString(),
+    };
+
+    if (zipResults.zipped) {
```
erunion, line 85: `zipResults` always here even if we aren't unzipping a ZIP? It doesn't look like this migrate command has a `--zip` argument.

erunion (approved):

src/commands/docs/migrate.ts
kanadgupta: yeah so we basically check in `attemptUnzip` to see if the main input argument is a zip file or not and if so, `zipResults.zipped` is `true`. it's a little wonky at the moment, very open to re-evaluating

src/lib/syncPagePath.ts
```diff
@@ -192,9 +211,9 @@ function sortFiles(files: PageMetadata<PageRepresentation>[]): PageMetadata<Page
  * and syncs those (either via POST or PATCH) to ReadMe.
  * @returns An array of objects with the results
  */
-export default async function syncPagePath(this: APIv2PageCommands) {
+export default async function syncPagePath(this: DocsUploadCommand) {
```
kanadgupta, line 222: whoops

src/index.ts
```diff
@@ -41,7 +41,7 @@ export const COMMANDS = {
   'openapi:validate': OpenAPIValidateCommand,
 
   whoami: WhoAmICommand,
-};
+} as const;
```
kanadgupta, line 44: sure why not!

PR #1232, opened 2025-04-24
docs: migration guide + legacy doc cleanup

kanadgupta:
## 🧰 Changes

@mjcuva pointed out a few gaps in our documentation as it currently exists:

- [x] some of these migration guide links were improperly pointing to the `next` branch when they should be pointing to the `v10` tag
- [x] the migration guide mentions `docs upload` which isn't technically out yet (i'll plan to add this back in https://github.com/readmeio/rdme/pull/1166
- [x] made a few tweaks to the callout language in the legacy doc so it's even clearer what these examples are doing.

## 🧬 QA & Testing

are you happy with these changes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ With `rdme`, you can manage your API definition (we support [OpenAPI](https://sp
 Not using ReadMe for your docs? No worries. `rdme` has a variety of tools to help you identify issues with your API definition — no ReadMe account required.
 
 > [!NOTE]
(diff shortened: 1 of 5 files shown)


erunion (approved):

mjcuva (approved):
Thanks for doing this! This is a lot clearer I think now

kanadgupta:
@mjcuva agreed, tysm for the feedback!

PR #1233, opened 2025-04-25
refactor: consolidate doc frontmatter validation logic

kanadgupta:
## 🧰 Changes

i'm about to open up another PR for RM-12535 that adds a new `-confirm-autofixes` flag, and as part of that i got fed up and totally DRY'd up our duplicated validation logic so it's now basically the same between `docs:upload` and `docs:migrate`.

i also added a tiny enhancement to our frontmatter autofixing logic that allows the `hidden` attribute to accept flags like `'true'` and `'false'`.

## 🧬 QA & Testing

`docs:migrate` is still in dire need of more test coverage so i've been testing it locally and confirmed that it's still working as expected. `docs:upload` has great test coverage so if those tests pass we should be in good shape.

### DIFF
--- a/src/lib/validateFrontmatter.ts
+++ b/src/lib/validateFrontmatter.ts
@@ -36,7 +36,9 @@ export async function validateFrontmatter(
   const filesWithFixableIssues = filesWithIssues.filter(result => result.fixableErrorCount);
   const filesWithUnfixableIssues = filesWithIssues.filter(result => result.unfixableErrors.length);
 
-  if (filesWithIssues.length) {
+  if (!filesWithIssues.length) {
+    validationSpinner.succeed(`${validationSpinner.text} no issues found!`);
+  } else {
     validationSpinner.warn(`${validationSpinner.text} issues found in ${filesWithIssues.length} file(s).`);
     if (filesWithFixableIssues.length) {
       if (isCI()) {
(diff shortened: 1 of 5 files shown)


src/lib/validateFrontmatter.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.debug(`validating frontmatter for ${file.filePath}`);
+    return fix.call(this, file.data, schema, mappings);
+  });
+
+  const filesWithIssues = validationResults.filter(result => result.hasIssues);
+  this.debug(`found ${filesWithIssues.length} files with issues: ${JSON.stringify(filesWithIssues)}`);
+  const filesWithFixableIssues = filesWithIssues.filter(result => result.fixableErrorCount);
+  const filesWithUnfixableIssues = filesWithIssues.filter(result => result.unfixableErrors.length);
+
+  if (filesWithIssues.length) {
```
erunion, line 39: Since the bulk of this method is in this block, how do you feel about flipping it?

```suggestion
  if (!filesWithIssues.length) {
    validationSpinner.succeed(`${validationSpinner.text} no issues found!`);
    return pagesToReturn
  }
```

erunion (approved):

src/lib/validateFrontmatter.ts
kanadgupta: done! https://github.com/readmeio/rdme/pull/1233/commits/bb79f25843457e999d701f139670f901c83b93bb

PR #1234, opened 2025-04-25
feat(docs): add ability to bypass autofix prompt

kanadgupta:
## 🧰 Changes

adds a new flag called `--confirm-autofixes` so users of `docs:upload` and `docs:migrate` can bypass the prompt to autofix any frontmatter issues. will be using this flag in our internal plugin work.

## 🧬 QA & Testing

added some more test coverage. does it all pass?

### DIFF
--- a/__tests__/commands/docs/upload.test.ts
+++ b/__tests__/commands/docs/upload.test.ts
@@ -362,7 +362,7 @@ describe('rdme docs upload', () => {
         projectsMeMock.done();
       });
 
-      it.only('should bypass prompt if `--confirm-autofixes` flag is passed', async () => {
+      it('should bypass prompt if `--confirm-autofixes` flag is passed', async () => {
         const getMock = getAPIv2MockForGHA({ authorization }).get('/versions/stable/guides/legacy-category').reply(404);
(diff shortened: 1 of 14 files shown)


__tests__/commands/docs/upload.test.ts
```diff
@@ -322,7 +322,7 @@ describe('rdme docs upload', () => {
       afterEach(githubActionsEnv.after);
 
       it('should create a guides page in ReadMe and include `x-readme-source-url` source header', async () => {
-        const headMock = getAPIv2MockForGHA({ authorization }).get('/versions/stable/guides/new-doc').reply(404);
```
erunion, line 325: i wonder if moving these calls from HEAD to GET is what is causing the slowdowns we've been seeing during upload.

erunion (approved):

__tests__/commands/docs/upload.test.ts
kanadgupta: maybe? i'm pretty sure it hits the same code path since our server doesn't have `HEAD` endpoints defined for guides

PR #1236, opened 2025-04-28
feat: better guides types, expose page stat type for `renderable` status

kanadgupta:
| 🚥 Resolves RM-12569 |
| :------------------- |

## 🧰 Changes

one half of the changes required for RM-12569 — the other half is in the internal plugin.

- [x] refactors our types a bit since our guides representation differs slightly for request vs response payloads (and exposing both in our exported `src/utils.ts` types
- [x] adds a new `renderable` type to our page stat results type

## 🧬 QA & Testing

mostly type changes! if types pass we should be in good shape.

### DIFF
--- a/__tests__/lib/apiError.test.ts
+++ b/__tests__/lib/apiError.test.ts
@@ -16,7 +16,7 @@ const response = {
   ],
 };
 
-describe('APIv1Error', () => {
+describe('#APIv1Error', () => {
   it('should handle ReadMe API errors', () => {
     const error = new APIv1Error(response);
 
--- a/src/lib/hooks/exported.ts
+++ b/src/lib/hooks/exported.ts
@@ -1,6 +1,7 @@
 import type { PageMetadata } from '../readPage.js';
 import type { PushResult } from '../syncPagePath.js';
 import type { Hooks } from '@oclif/core/interfaces';
+import type { GuidesResponseRepresentation } from '../types/index.js';
 
 interface MarkdownFileScanResultOptsBase {
   // required for the oclif hook types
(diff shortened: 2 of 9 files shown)


__tests__/lib/apiError.test.ts
```diff
@@ -16,7 +16,7 @@ const response = {
   ],
 };
 
-describe('APIv1Error', () => {
+describe('#APIv1Error', () => {
```
kanadgupta, line 19: unrelated lint warning 🫠

src/lib/hooks/exported.ts
```diff
@@ -49,22 +50,29 @@ export interface FailedUploadStatus extends BaseUploadStatus {
...
-export interface CreateUploadStatus extends BaseUploadStatus {
-  status: 'created';
+interface BaseSuccessUploadStatus extends BaseUploadStatus {
+  /**
+   * Upon upload, there a validation to determine if the page content is renderable or not.
+   * This is the result of that validation.
+   *
+   * - `status`: A flag for if the resource is renderable or not.
+   * - `error`: The rendering error. This is only present if the validation failed.
+   * - `message`: Additional details about the rendering error. This is only present if the validation failed.
```
kanadgupta, line 60: would be rad if these could be derived from the `json-schema-to-ts` type, but alas

__tests__/lib/apiError.test.ts
erunion: what was the warning?

erunion (approved):

__tests__/lib/apiError.test.ts
kanadgupta: CleanShot 2025-04-28 at 17 37 36@2x

kanadgupta: not opposed to disabling this rule

erunion: ah weird, it wants you to do `describe(APIv1Error, () => ...` and it could do `.constructor.name`.

that's... weird. i'm on board with disabling that if you want to PR it to our core standards

kanadgupta: annoyingly, if i pass the class itself, i get this lint warning:

CleanShot 2025-04-28 at 17 41 00@2x

erunion: yeah lets just disable `prefer-describe-function-title`. using a function as a test name is **funky**

PR #1241, opened 2025-05-12
fix: problem with `FormData` compositions on Node 24

erunion:
| 🚥 Resolves CX-1950 |
| :------------------- |

## 🧰 Changes

Undici changed something with how streams are processed in `FormData` payloads in the version that ships with Node 24 that has broken OpenAPI definition uploads within rdme there. Because we don't _really_ need to send these API definitions as a stream in our `multipart/form-data` payload to our public API I've changed them to just send it as a regular `File` object.

## 🧬 QA & Testing

Tested this locally on Node 22 and 24 and everything is green.

### DIFF
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -177,15 +177,15 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
       const { path } = await tmpFile({ prefix: 'rdme-openapi-', postfix: fileExtension });
       this.debug(`creating temporary file at ${path}`);
       fs.writeFileSync(path, specToUpload);
-      const stream = fs.createReadStream(path);
 
-      this.debug('file and stream created, streaming into form data payload');
-      body.append('schema', {
-        [Symbol.toStringTag]: 'File',
-        name: filename,
-        stream: () => stream,
-        type: isYaml ? 'application/x-yaml' : 'application/json',
-      });
+      this.debug('temporary file created, processing into form data payload');
+      body.append(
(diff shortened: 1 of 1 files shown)


src/commands/openapi/upload.ts
```diff
@@ -177,15 +177,15 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
...
-        [Symbol.toStringTag]: 'File',
-        name: filename,
-        stream: () => stream,
-        type: isYaml ? 'application/x-yaml' : 'application/json',
-      });
+
+      this.debug('temporary file created, processing into form data payload');
+      body.append(
+        'schema',
+        new File([fs.readFileSync(path)], filename, {
```
erunion, line 184: Can honestly probably just do this and not deal with the temp file creation.

```suggestion
        new File([specToUpload], filename, {
```

kanadgupta: not opposed, will defer to you here! this was a relic of our previous "stream spec to the registry and then pass the registry UUID to the upload endpoint" song and dance

kanadgupta (approved):
oh damn didn't realize the fix would be this straightforward. are we worried about any performance trade-offs with this change? i know that's why we moved to streaming in the first place but i don't think we ever determined if that even helped us much.

erunion:
> are we worried about any performance trade-offs with this change? i know that's why we moved to streaming in the first place but i don't think we ever determined if that even helped us much.

yeah we unfortunately don't really have any metrics on if that helped or not

PR #1242, opened 2025-05-12
fix(v9/openapi): problem with `FormData` compositions on Node 24

kanadgupta:
| 🚥 Resolves CX-1950 |
| :------------------- |

## 🧰 Changes

backports https://github.com/readmeio/rdme/pull/1241 over to the `v9` channel.

### DIFF
--- a/src/lib/streamSpecToRegistry.ts
+++ b/src/lib/streamSpecToRegistry.ts
@@ -23,18 +23,10 @@ export default async function streamSpecToRegistry(
   const { path } = await tmpFile({ prefix: 'rdme-openapi-', postfix: '.json' });
   debug(`creating temporary file at ${path}`);
   await fs.writeFileSync(path, spec);
-  const stream = fs.createReadStream(path);
 
-  debug('file and stream created, streaming into form data payload');
+  debug('temporary file, created, constructed form data');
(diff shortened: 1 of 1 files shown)


src/lib/streamSpecToRegistry.ts
```diff
@@ -23,18 +23,10 @@ export default async function streamSpecToRegistry(
...
   const formData = new FormData();
-  formData.append('spec', {
-    type: 'application/json',
-    name: 'openapi.json',
-    [Symbol.toStringTag]: 'File',
-    stream() {
-      return stream;
-    },
-  });
+  formData.append('spec', new File([fs.readFileSync(path)], path, { type: 'application/json' }), path);
```
kanadgupta, line 29: in `rdme@9` and APIv1, we don't need to do any converting back to YAML like we do in `rdme@10` since there's no git representation that we're trying to preserve.

kanadgupta: also i'm leaving the temporary file generation in place for now, imo it's handy way to debug what the bundling result is (and in hindsight maybe we should have kept it in place in https://github.com/readmeio/rdme/pull/1241 oops)

erunion (approved):

PR #1243, opened 2025-05-13
docs: convert legacy docs to MDX syntax

kanadgupta:
## 🧰 Changes

MDX syntax doesn't support HTML comments (and wow prettier does not play nicely with MDX) 😮‍💨 this converts our synced markdown pages over to MDX


erunion (approved):

runnabro (approved):
ty! hope you don’t mind I pushed a commit with some more fixes to those img tags <3

PR #1257, opened 2025-06-04
feat: properly display file size feature in openapi inspect command

olehshh:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

This PR introduces handling for the new analyzer features — specifically `rawFileSize` and `dereferencedFileSize` — as introduced in [readmeio/oas#982](https://github.com/readmeio/oas/pull/982).


__tests__/commands/openapi/__snapshots__/inspect.test.ts.snap
```diff
@@ -283,8 +285,10 @@ exports[`rdme openapi inspect > full reports > should generate a report for @rea
 exports[`rdme openapi inspect > full reports > should generate a report for @readme/oas-examples/3.0/json/readme.json 1`] = `
 "Here are some interesting things we found in your API definition. 🕵️
 
+ · Your dereferenced file size is 0.32 MB.
```
erunion, line 288: Unfortunate that me alphabetizing this object in `oas` caused this and raw file size to get dropped where they are in this report instead of at the end after the security type callout.

Wonder if we should split `dereferencedFileSize` and `rawFileSize` out of `general` and into a `file` or `buffer` query? Or if you want to manually sort them in here instead, but honestly it's probably fine as-is.

erunion (approved):
looks great!

__tests__/commands/openapi/__snapshots__/inspect.test.ts.snap
olehshh: I’d say I’d go ahead and use it for now, and if it turns out to cause any extra hassle, I’ll open a PR to improve it🤔

PR #1261, opened 2025-06-12
feat: new `reference upload` command

emilyskuo:
| 🚥 Resolves RM-12965 |
| :------------------- |

## 🧰 Changes

Creating a new `reference upload` command for users to upload MD pages to the reference section. This is entirely based on the `docs upload` command, with most of it being exactly the same, minus the route and some typing.

I had looked into making this an alias of `docs upload`, but oclif doesn't tell you how the command was invoked, and we'd have to parse `argv`s to figure out if docs or reference, which could potentially work but would also be pretty brittle. Otherwise, we could create a general upload command for markdown pages and add a type arg, which would be changing up behavior even more.

## 🧬 QA & Testing

@kanadgupta consolidated our markdown upload command tests in https://github.com/readmeio/rdme/pull/1262 🎉


__tests__/__fixtures__/docs/existing-docs/subdir/another-doc.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: This is another document title
 category:
-  uri: /branches/stable/categories/guides/category-slug
+  uri: category-slug
```
emilyskuo, line 4: Updating so we can use it with `references upload` as well

src/lib/types/openapiDoc.ts
```diff
@@ -6525,7 +6525,7 @@ const document = {
                         description: 'The endpoint HTTP method.',
                       },
                       path: { type: 'string', description: 'The endpoint path.' },
-                      schema: { nullable: true },
+                      schema: { type: 'string', nullable: true },
```
emilyskuo, line 6528: The lack of type here was breaking frontmatter validation with the error "nullable cannot be used without type" from ajv

erunion: 😬 I actually had to fix something similar for this over as part of my backgrounding work in https://github.com/readmeio/readme/pull/14881/commits/b09cb92c36831a1ca5ccc18fa035f8f663fc5319.

Since `type: string` is incorrect for this data shape, as it's a nullable object. Does `type: object` pass AJV here instead? Or if that doesn't work what about `additionalProperties: {}`?

```suggestion
                      schema: { type: 'object', properties: {}, nullable: true },
```

```suggestion
                      schema: { additionalProperties: {}, nullable: true },
```

Really wish I had just returned the schema as a string...

emilyskuo: `schema: { type: 'object', properties: {}, nullable: true },` works!

erunion: awesome, i'll see if i can update our OAS generation to make that change there so we don't have to worry about preserving it just here.

kanadgupta (changes requested):
this all looks great! the two blocking pieces of feedback should be fairly trivial: 
1. separating out the type fixes in the `src/lib/types/openapiDoc.ts` into its own PR
2. renaming the command from `rdme references upload` to `rdme reference upload`. could you also rename the two parent folders from "references" to "reference"?

src/lib/types/openapiDoc.ts
kanadgupta: ope, just seeing this. yeah if we could defer to the auto-generated OAS as our source of truth, i'd prefer that approach. ([more info in this comment](https://github.com/readmeio/rdme/pull/1261#discussion_r2151158098))

__tests__/__fixtures__/docs/existing-docs/subdir/another-doc.md
kanadgupta: update: worked on this in https://github.com/readmeio/rdme/pull/1262!

__tests__/commands/references/upload.test.ts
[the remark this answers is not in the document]
kanadgupta: update: figured this would help us move a little faster on support for custom pages, so i cranked this out in https://github.com/readmeio/rdme/pull/1262!

__tests__/commands/references/upload.test.ts
```diff
@@ -0,0 +1,822 @@
...
+
+import Command from '../../../src/commands/references/upload.js';
+import { getAPIv1Mock, getAPIv2Mock, getAPIv2MockForGHA } from '../../helpers/get-api-mock.js';
+import { githubActionsEnv } from '../../helpers/git-mock.js';
+import { runCommand, type OclifOutput } from '../../helpers/oclif.js';
+
+const key = 'rdme_123';
+const authorization = `Bearer ${key}`;
+
+describe('rdme references upload', () => {
```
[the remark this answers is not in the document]
kanadgupta: never mind, just merge this in instead: https://github.com/readmeio/rdme/pull/1262

src/lib/types/openapiDoc.ts
```diff
@@ -6525,7 +6525,7 @@ const document = {
                         description: 'The endpoint HTTP method.',
                       },
                       path: { type: 'string', description: 'The endpoint path.' },
-                      schema: { nullable: true },
+                      schema: { type: 'object', properties: {}, nullable: true },
```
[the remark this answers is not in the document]
emilyskuo: Gonna leave this as is as discussed during our sync

src/commands/reference/upload.ts
```diff
@@ -0,0 +1,89 @@
...
+      command: '<%= config.bin %> <%= command.id %> [path] --dry-run',
+    },
+  ];
+
+  static flags = {
+    key: keyFlag,
+    ...branchFlag(),
+    'confirm-autofixes': confirmAutofixesFlag,
+    'dry-run': Flags.boolean({
+      description: 'Runs the command without creating nor updating any References in ReadMe. Useful for debugging.',
```
kanadgupta, line 56:
```suggestion
      description: 'Runs the command without creating nor updating any reference pages in ReadMe. Useful for debugging.',
```

kanadgupta (approved):
one tiny copy edit below, otherwise LGTM! great work here ❤️‍🔥

PR #1265, opened 2025-06-18
feat: new `custompages upload` command

emilyskuo:
| 🚥 Resolves RM-12966 |
| :------------------- |

## 🧰 Changes

Adding a new `custompages upload` command for folks to upload MD or HTML custom pages. Largely copied from docs/reference upload with some minor tweaks (adding HTML as a valid input).

## 🧬 QA & Testing

See `__tests__/commands/custompages/upload.test.ts`. Also did some manual testing locally with some of the custom page test fixtures

### DIFF
--- a/src/lib/syncPagePath.ts
+++ b/src/lib/syncPagePath.ts
@@ -21,16 +21,14 @@ import { allowedMarkdownExtensions, findPages, type PageMetadata } from './readP
 import { categoryUriRegexPattern, parentUriRegexPattern } from './types/index.js';
 import { validateFrontmatter } from './validateFrontmatter.js';
 
-type PageRequestRepresentation =
-  | CustomPagesRequestRepresentation
-  | GuidesRequestRepresentation
-  | ReferenceRequestRepresentation;
+type GuidesReferenceRequestRepresentation = GuidesRequestRepresentation | ReferenceRequestRepresentation;
+
+type PageRequestRepresentation = CustomPagesRequestRepresentation | GuidesReferenceRequestRepresentation;
+
 type PageResponseRepresentation =
   | CustomPagesResponseRepresentation['data']
   | GuidesResponseRepresentation['data']
   | ReferenceResponseRepresentation['data'];
-
-type GuidesReferenceRequestRepresentation = GuidesRequestRepresentation | ReferenceRequestRepresentation;
 
 interface BasePushResult {
   filePath: string;
@@ -134,7 +132,7 @@ async function pushPage(
     }
 
     if (this.route === 'custom_pages') {
-      const customPagePayload = { ...payload } as CustomPagesRequestRepresentation;
+      const customPagePayload = structuredClone(payload) as CustomPagesRequestRepresentation;
       const type = path.extname(filePath).toLowerCase() === '.html' ? 'html' : 'markdown';
       if (typeof customPagePayload.content === 'object' && customPagePayload.content) {
         customPagePayload.content.type = type;
@@ -298,9 +296,10 @@ export default async function syncPagePath(this: APIv2PageUploadCommands) {
   const count = { succeeded: 0, failed: 0 };
 
   // topological sort the files
-  const sortedFiles = sortFiles(
-    (unsortedFiles as PageMetadata<GuidesReferenceRequestRepresentation>[]).sort(byParentPage),
-  );
+  const sortedFiles =
+    this.route === 'custom_pages'
+      ? (unsortedFiles as PageMetadata<CustomPagesRequestRepresentation>[])
+      : sortFiles((unsortedFiles as PageMetadata<GuidesReferenceRequestRepresentation>[]).sort(byParentPage));
 
   // push the files to ReadMe
   const rawResults: PromiseSettledResult<PushResult>[] = [];


src/commands/custompages/upload.ts
```diff
@@ -0,0 +1,92 @@
...
+import syncPagePath, {
+  type FailedPushResult,
+  type SkippedPushResult,
+  type CreatePushResult,
+  type UpdatePushResult,
+} from '../../lib/syncPagePath.js';
+
+const alphaNotice = 'This command is in an experimental alpha and is likely to change. Use at your own risk!';
+
+export default class CustomPgsUploadCommand extends BaseCommand<typeof CustomPgsUploadCommand> {
```
kanadgupta, line 14: nit: mind renaming this var everywhere?

```suggestion
export default class CustomPagesUploadCommand extends BaseCommand<typeof CustomPgsUploadCommand> {
```

src/lib/syncPagePath.ts
```diff
@@ -238,7 +238,12 @@ export default async function syncPagePath(this: APIv2PageUploadCommands) {
     );
   }
 
-  let unsortedFiles = await findPages.call(this, pathInput);
+  const validFileExtensions = ['.markdown', '.md', '.mdx'];
```
kanadgupta, line 241: could we instead reuse [this array](https://github.com/readmeio/rdme/blob/5c1567eb8d1155675ef7645bcce54e8f8c2e191c/src/lib/readPage.ts#L83) (i.e., create an array and import it in a few places) and import it here?

kanadgupta (commented):
a couple small nits below but overall this looks great! quick question since the answer is not immediately jumping out to me — if a user passes an HTML file into this, do we set the `type: 'html'`?

erunion (approved):
+1'ing kanad's feedback but this lgtm

src/lib/syncPagePath.ts
kanadgupta: perfect, tysm!

emilyskuo:
> quick question since the answer is not immediately jumping out to me — if a user passes an HTML file into this, do we set the type: 'html'?

I tested this out with an HTML file with just a title in the metadata and no we're not 😬 It uploaded as `type: 'markdown'` (I assume as a default). AJV doesn't catch it as an error, but we could add a check to `fix` to update `content.type` if it detects an html file

Screenshot 2025-06-24 at 11 50 58 AM

kanadgupta:
@emilyskuo hmm yeah there are a few places we could do this... i wouldn't necessarily consider this a "frontmatter fix" per se since this shouldn't be an attribute that users specify in their frontmatter — what if we did it around here instead:

https://github.com/readmeio/rdme/blob/ad93a7ad177c343d7a9d986ebf39a6568d65025f/src/lib/syncPagePath.ts#L100-L122

in those if-statements, we're patching the category/parent page data in the payload prior to hitting the API, so maybe that's also a good spot to do this as well? i believe this `pushPage` function should have access to all the info you need (the full file name, the fixed up frontmatter data, etc.)

let me know if you think of a better spot or if you run into any issues with this approach!

src/lib/syncPagePath.ts
```diff
@@ -1,24 +1,36 @@
...
+import { allowedMarkdownExtensions, findPages, type PageMetadata } from './readPage.js';
 import { categoryUriRegexPattern, parentUriRegexPattern } from './types/index.js';
 import { validateFrontmatter } from './validateFrontmatter.js';
 
-type PageRequestRepresentation = GuidesRequestRepresentation | ReferenceRequestRepresentation;
-type PageResponseRepresentation = GuidesResponseRepresentation['data'] | ReferenceResponseRepresentation['data'];
+type PageRequestRepresentation =
+  | CustomPagesRequestRepresentation
+  | GuidesRequestRepresentation
+  | ReferenceRequestRepresentation;
```
kanadgupta, line 27: nit: can you rearrange these types a bit so we can do something like this?

```suggestion
  | GuidesReferenceRequestRepresentation
```

src/lib/syncPagePath.ts
```diff
@@ -121,6 +133,17 @@ async function pushPage(
       }
     }
 
+    if (this.route === 'custom_pages') {
+      const customPagePayload = { ...payload } as CustomPagesRequestRepresentation;
```
kanadgupta, line 137: this feels a tad safer given how nested this object may be:

```suggestion
      const customPagePayload = structuredClone(payload) as CustomPagesRequestRepresentation;
```

src/lib/syncPagePath.ts
```diff
@@ -121,6 +133,17 @@ async function pushPage(
...
+      const type = path.extname(filePath).toLowerCase() === '.html' ? 'html' : 'markdown';
+      if (typeof customPagePayload.content === 'object' && customPagePayload.content) {
+        customPagePayload.content.type = type;
+      } else {
+        customPagePayload.content = { type };
+      }
```
kanadgupta, line 143: ohhh smart 🧠

src/lib/syncPagePath.ts
```diff
@@ -265,7 +298,9 @@ export default async function syncPagePath(this: APIv2PageUploadCommands) {
   const count = { succeeded: 0, failed: 0 };
 
   // topological sort the files
-  const sortedFiles = sortFiles((unsortedFiles as PageMetadata<PageRequestRepresentation>[]).sort(byParentPage));
+  const sortedFiles = sortFiles(
+    (unsortedFiles as PageMetadata<GuidesReferenceRequestRepresentation>[]).sort(byParentPage),
+  );
```
kanadgupta, line 303: can you do this instead ([derived from the work i did here](https://github.com/readmeio/rdme/pull/1272/files#diff-387a87252a58334e9a8eac46fabbe6a288d96043a28f79bf0f342d36ff826a7c))? that way it doesn't actually run the sort function for custom pages:

```suggestion
  const sortedFiles =
    this.route === 'custom_pages'
      ? (unsortedFiles as PageMetadata<CustomPagesRequestRepresentation>[])
      : sortFiles((unsortedFiles as PageMetadata<GuidesReferenceRequestRepresentation>[]).sort(byParentPage));
```

kanadgupta (changes requested):
just a handful of requested changes that should be quick to address, but on the whole this looks phenomenal. great work ❤️‍🔥

src/lib/syncPagePath.ts
emilyskuo: Took that idea from `fix` haha

kanadgupta (approved):
hell yeah brother 🤘🏽 this was a surprisingly gnarly change, awesome work!

PR #1268, opened 2025-06-23
refactor: smol type improvement for reference pages

kanadgupta:
## 🧰 Changes

just a small type addition to reflect the new support for reference pages. we'll want to make a similar change for custom pages + changelog too.


kanadgupta:
ignore the CI failure, working on a fix for that as we speak!

erunion (approved):

PR #1278, opened 2025-06-26
feat: add legacy id flag to openapi upload

emilyskuo:
| 🚥 Resolves RM-13049 |
| :------------------- |

## 🧰 Changes

Adding in a `legacy-id` flag to the `openapi upload` command to allow migrated users to match their specs in refactored to their legacy spec

## 🧬 QA & Testing

See updates to `openapi/upload.test.ts`.

Also tested this locally by:
 - Creating a legacy project
 - Uploading `petstore-simple-weird-version.json`
 - Migrating the project to Refactored (this needed some finagling in mongo, but I'm pretty confident that didn't mess anything up)
   - The slug in Refactored became `single-path.json`
 - Running `bin/dev.js openapi upload` with `--legacy-id` pulled from mongo
 - `single-path.json` got updated and we did not create a new `petstore-simple-weird-version.json` definition

### DIFF
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -1,5 +1,5 @@
 import type {
-  APIDefinitionData,
+  APIDefinitionsRepresentation,
   APIUploadSingleResponseRepresentation,
   APIUploadStatus,
   StagedAPIUploadResponseRepresentation,
@@ -159,11 +159,11 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
 
     const headers = new Headers({ authorization: `Bearer ${this.flags.key}` });
 
-    const existingAPIDefinitions = await this.readmeAPIFetch(`/branches/${branch}/apis`, { headers }).then(res =>
+    const existingAPIDefinitions = (await this.readmeAPIFetch(`/branches/${branch}/apis`, { headers }).then(res =>
       this.handleAPIRes(res),
-    );
+    )) as APIDefinitionsRepresentation;
 
-    const matchingAPIDefinition = existingAPIDefinitions?.data?.find((d: APIDefinitionData) => {
+    const matchingAPIDefinition = existingAPIDefinitions?.data?.find(d => {
       if (this.flags['legacy-id']) {
         return d?.legacy_id === this.flags['legacy-id'];
       }
--- a/src/lib/types/index.ts
+++ b/src/lib/types/index.ts
@@ -98,6 +98,4 @@ export type PageResponseSchema<T extends PageRoute> = FromSchema<
  * Derived from our API documentation, this is the schema for the get APIs response
  * as we receive it from the ReadMe API.
  */
-type APIDefinitionsRepresentation = FromSchema<apiDefinitionsSchema, { keepDefaultedPropertiesOptional: true }>;
-
-export type APIDefinitionData = APIDefinitionsRepresentation['data'][0];
+export type APIDefinitionsRepresentation = FromSchema<apiDefinitionsSchema, { keepDefaultedPropertiesOptional: true }>;


src/commands/openapi/upload.ts
```diff
@@ -122,7 +130,7 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
 
     let filename = specFileTypeIsUrl ? nodePath.basename(specPath) : slugify.default(specPath);
 
-    if (!specFileTypeIsUrl && filename !== specPath) {
+    if (!specFileTypeIsUrl && filename !== specPath && !this.flags['legacy-id']) {
```
emilyskuo, line 133: This warning isn't really relevant if you're using `legacy-id` since you can't use it with `slug`

src/commands/openapi/upload.ts
```diff
@@ -147,10 +155,20 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
       this.handleAPIRes(res),
     );
 
-    // if the current slug already exists, we'll use PUT to update it. otherwise, we'll use POST to create it
-    const method = existingAPIDefinitions?.data?.some((d: { filename: string }) => d.filename === filename)
-      ? 'PUT'
-      : 'POST';
+    const matchingAPIDefinition = existingAPIDefinitions?.data?.find((d: { filename: string; legacy_id: string }) => {
+      if (this.flags['legacy-id']) {
+        return d?.legacy_id === this.flags['legacy-id'];
```
emilyskuo, line 160: Do we want to throw an error if a user passes in `legacy-id` and we don't find a matching API definition?

src/commands/openapi/upload.ts
```diff
@@ -40,6 +40,14 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
         'If set, file overwrites will be made without a confirmation prompt. This flag can be a useful in automated environments where prompts cannot be responded to.',
       hidden: true,
     }),
+    'legacy-id': Flags.string({
+      summary: 'The legacy ID for your API definition.',
+      description:
+        'This is only used for legacy rdme CLI workflows and only applies if your project, and this API definition, predates ReadMe Refactored. We consider this value to be deprecated and recommend against relying on it going forward.',
```
erunion, line 46: could use some more massaging but i think we should point them towards `slug` being our recommended argument 

```suggestion
        'This is only used for legacy rdme CLI workflows and only applies if your project, and this API definition, predates ReadMe Refactored. We consider argument value to be deprecated and recommend against relying on it going forward, instead prefer using `slug`.',
```

src/commands/openapi/upload.ts
erunion: yeah we should probably throw an error

erunion (approved):

src/commands/openapi/upload.ts
kanadgupta: agreed! while you're here, mind also adding this? i don't think it makes sense to emit this warning if `--slug` is passed either

```suggestion
    if (!specFileTypeIsUrl && filename !== specPath && !this.flags['legacy-id'] && !this.flags['slug']) {
```

src/commands/openapi/upload.ts
kanadgupta: i'm not sure if the error should be thrown here (it might make sense to do it after this `.find` statement?) but agreed that we should throw an error if they pass the legacy flag and it doesn't match anything!

src/commands/openapi/upload.ts
kanadgupta: how about this?

```suggestion
        'This is only used for legacy `rdme` CLI workflows and only applies if your project, and this API definition, predate ReadMe Refactored. This flag is considered deprecated and we recommend using `--slug` instead.',
```

kanadgupta (approved):
looks great, tysm!

src/lib/types/index.ts
```diff
@@ -90,3 +93,11 @@ export type PageResponseSchema<T extends PageRoute> = FromSchema<
...
   { keepDefaultedPropertiesOptional: true }
 >;
+
+/**
+ * Derived from our API documentation, this is the schema for the get APIs response
+ * as we receive it from the ReadMe API.
+ */
+type APIDefinitionsRepresentation = FromSchema<apiDefinitionsSchema, { keepDefaultedPropertiesOptional: true }>;
+
+export type APIDefinitionData = APIDefinitionsRepresentation['data'][0];
```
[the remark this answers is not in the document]
emilyskuo: Ooooh okay thanks!

kanadgupta (approved):
no notes ✨

PR #1283, opened 2025-07-02
ci: moving from eslint to biome

erunion:
## 🧰 Changes

I've been staring down the barrel of upgrading our shared ESLint config to support ESLint 9 over in https://github.com/readmeio/standards/pull/993 and... I honestly would rather live my life instead. I've been really interested in https://biomejs.dev/ as of late and after playing around with it for not even 10 minutes I am sold.

* It's almost a fully drop-in replacement for ESLint
* I was able to configure and retain our opinionated import grouping ruleset[^1] without any trouble.
* It found and fixed issues that ESLint didn't, and can't, detect!
* It linted our codebase in half a second vs ESLint taking 4.

They're still working on supporting Markdown and YAML so we can't fully ditch Prettier too but as it is Biome has got the juice.

[^1]: https://github.com/readmeio/standards/blob/bc29338858c55da37631addb36582ec5feac11b0/packages/eslint-config/index.js#L25-L36

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -11,10 +11,8 @@
     "includes": [
       "**/*.{js,ts}",
 
-      // test result artifacts
+      // Biome's `vcs.useIgnoreFile` config should be able to pick these up but it doesn't.
       "!coverage/**",
-
-      // release build artifacts
       "!dist/**",
       "!dist-gha/**",
     ],


__tests__/commands/changelogs/index.test.ts
```diff
@@ -41,8 +43,8 @@ describe('rdme changelogs', () => {
   });
 
   describe('existing changelogs', () => {
-    let simpleDoc;
-    let anotherDoc;
+    let simpleDoc: PageObject;
```
erunion, line 46: It flagged that these didn't have an explicit type. I don't know where to put `page.types.js` but I didn't want to copy-paste this `PageObject` interface in a handful of files.

__tests__/helpers/vitest.matchers.ts
```diff
@@ -13,7 +13,7 @@ interface CustomMatchers<R = unknown> {
 }
 
 declare module 'vitest' {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  // biome-ignore lint/suspicious/noExplicitAny: This is the type for a custom matcher.
```
erunion, line 16: I **love** that it requires you give a reason why you're ignoring a rule.

__tests__/lib/createGHA.test.ts
```diff
@@ -86,7 +84,7 @@ describe('#createGHA', () => {
         expect(yamlOutput).toBeValidSchema(ghaWorkflowSchema);
         expect(yamlOutput).toMatchSnapshot();
         expect(fs.writeFileSync).toHaveBeenCalledWith(getGHAFileName(fileName), expect.any(String));
-        expect(console.info).toHaveBeenCalledTimes(1);
+        expect(consoleInfoSpy).toHaveBeenCalledTimes(1);
```
erunion, line 87: Moved this over to the spy because it was already available and used in other tests.

biome.jsonc
```diff
@@ -0,0 +1,98 @@
...
+    "ignoreUnknown": false,
+    "includes": [
+      "**/*.{js,ts}",
+
+      // test result artifacts
+      "!coverage/**",
+
+      // release build artifacts
+      "!dist/**",
+      "!dist-gha/**",
```
erunion, line 19: I thought with `vcs.useIgnoreFile: true` we could get rid of these and have it automatically exclude files in our `.gitignore` file but it didn't work.

src/commands/docs/migrate.ts
```diff
@@ -94,7 +94,7 @@ export default class DocsMigrateCommand extends BaseCommand<typeof DocsMigrateCo
     }
 
     validationHookResults.successes.forEach(success => {
-      if (success.result && success.result.pages.length) {
+      if (success.result?.pages.length) {
```
erunion, line 97: It recommended, but thankfully didn't autofix, that we move this to optional chaining.

src/commands/openapi/reduce.ts
```diff
@@ -131,8 +131,7 @@ export default class OpenAPIReduceCommand extends BaseCommand<typeof OpenAPIRedu
         choices: (prev, values) => {
           const paths: string[] = values.paths;
           let methods = paths
-            .map((p: string) => Object.keys(parsedPreparedSpec.paths?.[p] || {}))
-            .flat()
```
erunion, line 135: Blown away that it detected and autofixed this to a `.flatMap()`.

src/commands/openapi/resolve.ts
```diff
@@ -139,14 +136,13 @@ export default class OpenAPIResolveCommand extends BaseCommand<typeof OpenAPIRes
...
     circularRefs: string[],
   ): void {
-    // eslint-disable-next-line @typescript-eslint/no-this-alias
     const self = this;
     const createdRefs = new Set<string>();
 
     function replaceRef(schemaName: string, propertyName: string, refSchemaName: string) {
       const schema = schemas[schemaName];
-      if ('properties' in schema) {
-        schema.properties![propertyName] = { $ref: `#/components/schemas/${refSchemaName}` };
```
erunion, line 149: Rewrote this off of the non-null assertion because that was being picked up by their default ruleset.

Screenshot 2025-07-01 at 6 04 00 PM

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["biomejs.biome"]
```
erunion, line 2: So far this extension seems to work as well as the ESLint one.

src/lib/frontmatter.ts
```diff
@@ -1,8 +1,8 @@
-import type { Mappings } from './readmeAPIFetch.js';
```
erunion, line 1: Always bugged me that our ESLint import grouping config never grouped `type` imports like non-type imports. Really neat that Biome just does it all.

src/lib/hooks/prerun.ts
```diff
@@ -13,10 +13,8 @@ const hook: Hook.Prerun = async function run(options) {
   if (options.Command?.flags?.key) {
     this.debug('current command has --key flag');
     if (isTest()) {
-      // eslint-disable-next-line no-param-reassign
```
erunion, line 16: Not having a `no-param-reassign` rule might be an issue but with TS we've grown to be pretty good at not writing code that gets flagged with it.

src/lib/pageCommandProperties.ts
```diff
@@ -57,20 +57,20 @@ export function examples(section: Section) {
 
 export function baseFlags(section: Section) {
   const fileType = section === 'Custom Pages' ? 'Markdown/HTML' : 'Markdown';
-  let items;
```
erunion, line 60: It didn't like that `items` didn't have an explicit type.

src/lib/unzip.ts
```diff
@@ -30,7 +30,7 @@ export async function attemptUnzip(pathInput: string): Promise<MarkdownFileScanR
       const newWorkingDir = path.join(outputDir, path.basename(pathInput, '.zip'));
 
       return { pathInput: newWorkingDir, zipped: true as const, unzippedDir: newWorkingDir };
-    } catch (e) {
+    } catch (_e) {
```
erunion, line 33: Biome flags when we don't use variables but since we can't do `} catch () {` it says to prefix an unused variable with an underscore.

src/types.ts
```diff
@@ -1,23 +1,23 @@
 import type { PageRequestSchema, PageResponseSchema } from './lib/types/index.js';
 
 export type {
-  APIKeyRepresentation,
```
erunion, line 4: I'm full :sickos: mode over here at these imports being autofixed to be properly alphabetized now.

bin/write-gha-pjson.js
```diff
@@ -20,7 +21,6 @@ function writeGitHubActionsPackageJson() {
   // set the correct targets for GitHub Actions
   current.oclif.commands.target = newTarget;
   Object.values(current.oclif.hooks).forEach(hook => {
-    // eslint-disable-next-line no-param-reassign
```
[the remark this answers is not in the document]
erunion: Looks like it's part of the `project` domain but enabling `linter.domains.project: all` doesn't seem to enable it. Not sure why but I've manually turned it on and addressed a couple issues.

src/lib/pageCommandProperties.ts
erunion: Rolling this change back and adding an ignore w/ an explanation that the   switch statement gives us our types.

.vscode/settings.json
```diff
@@ -1,5 +1,10 @@
 {
+  "editor.codeActionsOnSave": {
+    "source.fixAll": "explicit"
+  },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
+  // explicitly disable ESLint to avoid conflicts with Biome
+  "eslint.enable": false,
```
kanadgupta, line 7: i added the changes in this file b/c i imagine there might be weird side effects when vscode is running eslint + biome simultaneously and i figured it's safer to be explicit here. we should probably do the same with prettier (i.e., explicitly disable the prettier extension and set the default formatter to biome) if/when we migrate off of that too.

biome.jsonc
```diff
@@ -0,0 +1,91 @@
...
+  },
+  "formatter": {
+    // We still use Prettier for formatting.
+    "enabled": false,
+  },
+  "linter": {
+    "enabled": true,
+    "domains": {
+      "project": "all",
+      "test": "all",
```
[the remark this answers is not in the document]
kanadgupta: ugh nvm i just played around locally and looks like we need this 😔

biome.jsonc
kanadgupta: this feels like a pretty rough oversight and doesn't inspire confidence tbh 😔 but glad the workaround is easy enough at least. mind adding a comment explaining this?

kanadgupta (approved):
i'm slightly bummed about some of the rules we're losing (the eslint ecosystem is very difficult to top in this regard) but i think the tradeoffs seem worth it — eslint has been giving us grief for ages and biome's speed + out-of-the-box capabilities (the fact that it plays much better with TS 🫰🏽) + simpler configuration are all very promising.

this is a very big workflow change and i think we should make sure the team is fully on board before we move forward with other repos (and let's also make sure the custom rules support is adequate for our needs) but i'm in favor of this! let's also think about [a shared config file that we publish](https://biomejs.dev/guides/big-projects/#exporting-a-biome-configuration-from-an-npm-package) from our [`standards`](https://github.com/readmeio/standards) repo at some point soon.

thanks for kicking off this effort ❤️‍🔥

PR #1286, opened 2025-07-02
chore: remove bidi checks for page uploads

emilyskuo:
| 🚥 Resolves RM-13222 |
| :------------------- |

## 🧰 Changes

We had added in a check for projects with bidi connections, and required them to use the `--skip-validation` flag if it was enabled in order to use rdme to upload pages. This was maybe a bit overly prescriptive and we would actually want that validation to happen, even if bidi is enabled. We haven't fully ironed out what the behavior should be for customers using both rdme and bidi to sync docs, but for now, we're gonna remove all guardrails and checks. Once customers start really using these flows, we can see if we need to make updates.

## 🧬 QA & Testing

Updated tests & they're still passing


kanadgupta (approved):
no notes!

PR #1307, opened 2025-07-15
chore(biome): incorporating our new common biome config

erunion:
## 🧰 Changes

This incorporates our new [@readme/standards](http://npm.im/@readme/standards) package for colocating our Prettier and Biome configs. Prettier was previously a part of `@readme/eslint-config` but as we're slowly beginning to move away from ESLint it makes sense to have this Prettier config decoupled from ESlint.

### DIFF
--- a/bin/json-schema-store.js
+++ b/bin/json-schema-store.js
@@ -55,4 +55,5 @@ async function refreshSchemas() {
   }
 }
 
-await refreshSchemas();
+// biome-ignore lint/nursery/noFloatingPromises: We use rollup to bundle this file and it doesn't play well with top-level await.
+refreshSchemas();
--- a/bin/json-schema-store.js
+++ b/bin/json-schema-store.js
@@ -55,5 +55,4 @@ async function refreshSchemas() {
   }
 }
 
-// biome-ignore lint/nursery/noFloatingPromises: We use rollup to bundle this file and it doesn't play well with top-level await.
-refreshSchemas();
+await refreshSchemas();
--- a/bin/json-schema-store.js
+++ b/bin/json-schema-store.js
@@ -55,4 +55,4 @@ async function refreshSchemas() {
   }
 }
 
-refreshSchemas();
+await refreshSchemas();
--- a/bin/run.js
+++ b/bin/run.js
@@ -16,4 +16,4 @@ async function main() {
   });
 }
 
-main();
+await main();
--- a/bin/run.js
+++ b/bin/run.js
@@ -16,4 +16,5 @@ async function main() {
   });
 }
 
-await main();
+// biome-ignore lint/nursery/noFloatingPromises: We use rollup to bundle this file and it doesn't play well with top-level await.
+main();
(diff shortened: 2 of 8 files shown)


bin/json-schema-store.js
```diff
@@ -55,4 +55,4 @@ async function refreshSchemas() {
   }
 }
 
-refreshSchemas();
+await refreshSchemas();
```
erunion, line 58: These were all floating promises that began to get flagged because we have `noFloatingPromises` enabled in our common config.

https://github.com/readmeio/standards/blob/92242ec1d1bf08bc1a0c3e4d6f340cdbb74397a6/packages/standards/biome.jsonc#L27-L31

biome.jsonc
```diff
@@ -22,19 +23,14 @@
     "enabled": false,
   },
   "linter": {
-    "enabled": true,
     "domains": {
+      // These domains are configured in our core standards package but Biome `extends` seems to
+      // not load it in so we have to redefine them.
```
erunion, line 28: Removing this `domains` config to rely on the one in the common config causes the `useImportExtensions` rule, which is enabled in `domains.project` to stop being loaded:

```bash
$ npx biome check
bin/set-version-output.js:6:1 suppressions/unused ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ Suppression comment has no effect. Remove the suppression or make sure you are suppressing the correct rule.
  
    4 │ import * as core from '@actions/core';
    5 │ 
  > 6 │ // biome-ignore lint/correctness/useImportExtensions: This file exists but Biome wants to use the `.d.ts` file instead.
      │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    7 │ import { getMajorPkgVersion, getNodeVersion } from '../dist/lib/getPkg.js';
    8 │ 
  

Checked 92 files in 47ms. No fixes applied.
Found 1 warning.
```

kanadgupta: thanks i hate it

biome.jsonc
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "https://biomejs.dev/schemas/2.1.1/schema.json",
   "root": true,
+  "extends": ["@readme/standards/biome"],
   "vcs": {
```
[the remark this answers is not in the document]
erunion: No, and I can get rid of `formatter` too.

bin/run.js
```diff
@@ -16,4 +16,4 @@ async function main() {
   });
 }
 
-main();
+await main();
```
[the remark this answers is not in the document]
erunion: oh my b

PR #1309, opened 2025-07-17
feat(openapi/upload): docs + DX enhancements for slugs vs. legacy IDs

kanadgupta:
## 🧰 Changes

folks have been a little confused with the slug vs. legacy id parameter paradigm, so this PR enhances the DX and docs for `rdme openapi upload` in a few ways:

- [x] expanded upon the whole legacy ID ➡️ slug change in the migration guide
- [x] added a little bit of smarter DX for when a user passes in a `--slug <some slug that looks eerily like a mongo object id>` (which is a no-no since object IDs should be using `--legacy-id`):
   - if a user passes an object ID and it matches one of the legacy IDs in the API definitions list, it'll prompt the user before updating that API definition, and emit a warning telling them what the correct slug should be. if this is run in CI, it'll throw an error that tells them the correct file name.
     CleanShot 2025-07-17 at 17 28 43@2x

   - if the object ID doesn't match anything, it'll proceed as normal but emit a warning telling them to check out our migration guide, just in case.
     CleanShot 2025-07-17 at 17 31 55@2x

- [x] if a user passes the `--legacy-id` flag and there aren't any matches but there are legacy API definitions in the project, the error message will suggest those
- [x] if a user passes the `--legacy-id` flag and there is a match, the warning is now a bit more custom tailored and will inform them of what the slug is and encourage them to use that instead

## 🧬 QA & Testing

wrote a bunch of tests for every permutation of this logic, and confirmed it works against my lil clone of our internal handbook.


erunion (approved):
no notes

src/commands/openapi/upload.ts
```diff
@@ -170,9 +221,21 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
 
     if (this.flags['legacy-id']) {
       if (!matchingAPIDefinition) {
-        throw new Error(`No API definition found with legacy ID ${this.flags['legacy-id']}.`);
+        let existingDefinitionSuggestion = ' Did you mean to use the `--slug` flag instead?';
+        if (existingAPIDefinitions.length === 1) {
+          existingDefinitionSuggestion = ` 1 file was found — try replacing \`--legacy-id ${this.flags['legacy-id']}\` with \`--slug ${existingAPIDefinitions[0].filename}\`.`;
+        } else if (existingAPIDefinitions.length > 1) {
+          existingDefinitionSuggestion = ` ${existingAPIDefinitions.length} file(s) were found — try passing one of them via the \`--slug\` flag instead: ${filenames}.`;
+        }
```
emilyskuo, line 229: Not really blocking, but this behavior feels a little odd to me:
 - Person uploads a spec with `--legacy-id [ID-lookin-thing]`
 - We don't find a matching spec to that ID
 - If no specs exist, we suggest they upload using `--slug=[ID-lookin-thing]` ?
 - If spec(s) are found, we suggest they overwrite a different file with an unrelated slug

Because we don't know if any of the existing specs are actually the same one or not, so it feels a little strong to suggest that they try one of those slugs. Idk what I'd do instead though 😰

PR #1312, opened 2025-07-28
refactor(openapi): cleaner approach to `prepareOas` helper

kanadgupta:
## 🧰 Changes

as part of the work in https://github.com/readmeio/rdme/pull/1313, this PR refactors our `prepareOas` helper (and by extension, the entire `openapi` family of commands) to be a little bit more conformant to the oclif command class pattern. this will clean up our debug logs a bit and should make it a bit easier to add new `openapi` commands going forward.

## 🧬 QA & Testing

no end user changes — if tests continue to pass we should be in good shape!

### DIFF
--- a/src/commands/openapi/resolve.ts
+++ b/src/commands/openapi/resolve.ts
@@ -22,6 +22,8 @@ type Schema = OpenAPIV31.ReferenceObject | OpenAPIV31.SchemaObject;
 type SchemaCollection = Record<string, Schema>;
 
 export default class OpenAPIResolveCommand extends BaseCommand<typeof OpenAPIResolveCommand> {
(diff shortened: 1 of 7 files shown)


src/commands/openapi/resolve.ts
```diff
@@ -50,6 +52,7 @@ export default class OpenAPIResolveCommand extends BaseCommand<typeof OpenAPIRes
 
   static flags = {
     out: Flags.string({ description: 'Output file path to write resolved file to' }),
+    title: titleFlag,
```
erunion, line 55: What's the resolve command doing with this?

erunion (approved):

Issue #1331, opened 2025-08-27
Required Node version is wrong


svaccari:
Hi,

`rdme@10` user here.
If I force Node to v14 I get this:

```
> nvm use 14
Now using Node v14.20.0 (npm 6.14.17)
> rdme
(node:4619) Warning: Node version must be >=18.0.0 to use this CLI. Current node version: 14.20.0
```

But if I actually use v18 I get following error:

```
> nvm use 18
Now using Node v18.20.8 (npm 10.8.2)
> rdme openapi upload swagger.json --branch=test --key=*** --slug=test
Updating Readme branch: test

File is not defined
```

`File is not defined` error is due to Node version.
All good with Node v20.

PR #1333, opened 2025-08-28
fix(oclif): disable engine warning

kanadgupta:
| 🚥 Resolves https://github.com/readmeio/rdme/issues/1331 |
| :------------------- |

## 🧰 Changes

as first discovered in https://github.com/readmeio/rdme/issues/1331, [`@oclif/core` uses **_its own_** node engine requirements and emits a warning](https://github.com/oclif/core/blob/70d3f192862a5adb548cbda68c6ee1ca8f724110/src/index.ts#L12-L27) if its node version requirements aren't met:

they require v18 or later but we require v20 or later, so the warning looks like this (which is wrong):

```
(node:4619) Warning: Node version must be >=18.0.0 to use this CLI. Current node version: 14.20.0
```

due to this being misleading, we're going to disable this warning entirely. i don't **_think_** we need to emit warnings for mismatched versions of node.js (since the warning is already emitted when users install this CLI via `npm`) but i'm open to reevaluating that.

## 🧬 QA & Testing

confirmed that using a version of node.js less than 18 no longer emits this warning.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -3,6 +3,13 @@
 import stringArgv from 'string-argv';
 
 async function main() {
+  /**
+   * Disables the oclif engine warning. For some reason the versions that are flagged are totally separate from our engines requirements.
+   *
+   * @see {@link https://github.com/oclif/core/blob/70d3f192862a5adb548cbda68c6ee1ca8f724110/src/index.ts#L12}
+   */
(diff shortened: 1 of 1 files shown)


erunion (approved):
insane that they do this

PR #1344, opened 2025-09-16
fix(openapi/upload): handle funky URLs

kanadgupta:
| 🚥 Resolves RM-13895 |
| :------------------- |

## 🧰 Changes

fixes a few issues where URLs with no path (e.g., `https://example.com`) or URLs with query params (e.g., `https://example.com/openapi?something=true`) weren't being handled properly.

## 🧬 QA & Testing

do tests pass?


emilyskuo (approved):
so funky

PR #1346, opened 2025-09-16
chore(dev): env var for pointing to localhost

kanadgupta:
## 🧰 Changes

@domharrington had inquired a while back about how to point to your local environment via `rdme`. there hasn't been a great way other than manually editing the source and rebuilding the CLI yourself. this PR fixes that!


domharrington:
Thank you 😄

Issue #1350, opened 2025-09-26
`openapi upload` no longer accepts `--workingDirectory` option


kitemongerer:
We're trying to migrate to readme v10 but we need the `--workingDirectory` param for `openapi upload`. It used to be available but is now only present in the `openapi validate` command. This means our multi-file spec can pass validate but can't be uploaded

chrisminett:
I managed to work around this by switching to the needed working directory, and then running the rdme bin directly with node:

```sh
cd openapi
node ../node_modules/rdme/bin/run.js openapi upload --key=...
```

That seemed to work, although it has revealed other issues from this forced upgrade (no way to select the target to update, duplicate pages added, etc.)

PR #1369, opened 2025-10-08
fix(openapi/upload): restore `--working-directory`

kanadgupta:
| 🚥 Resolves https://github.com/readmeio/rdme/issues/1350, CX-2286 |
| :------------------- |

## 🧰 Changes

restores the `--working-directory` (formerly known as `--workingDirectory`) flag for `openapi upload`. more chatter about this can be found [here](https://linear.app/readme-io/issue/CX-2286/cli-openapi-upload-command-not-respecting-relative-working-directory#comment-ee239f54).

## 🧬 QA & Testing

do tests pass?


src/commands/openapi/convert.ts
```diff
@@ -28,7 +28,7 @@ export default class OpenAPIConvertCommand extends BaseCommand<typeof OpenAPICon
   static flags = {
     out: Flags.string({ description: 'Output file path to write converted file to' }),
     title: titleFlag,
-    workingDirectory: workingDirectoryFlag,
+    'working-directory': workingDirectoryFlag,
```
erunion, line 31: Since this rename feels like a breaking change should we support workingDirectory and working-directory but mark the former as deprecated?

erunion: nevermind just saw the alias!

erunion (approved):

src/commands/openapi/convert.ts
kanadgupta: yep you got it! `--workingDirectory` will still work but using it will now emit a deprecation notice

PR #1437, opened 2026-04-17
fix: prevent duplicate page creation during docs upload

jesseyowell:
## Summary

Addresses an issue where pages could be silently duplicated with `-0`/`-1` suffixed slugs during upload. This happened when a transient GET failure (5xx/429) caused the client to mistakenly POST a new page for one that already existed, and the server would silently create it with a suffixed slug rather than rejecting it.

### Changes to `src/lib/syncPagePath.ts`

- **Retry logic for transient GET failures**: The initial GET check (to determine create vs. update) now retries up to 3 times with exponential backoff for 429 and 5xx responses. Previously, a transient server error could be misinterpreted as a 404, leading to a duplicate POST on re-run.
- **409 Conflict handling on POST**: If the server returns a 409 Conflict during page creation (indicating the page already exists), the client now falls back to a PATCH (update) instead of failing or allowing the server to create a duplicate with a `-1` suffix.
- **Slug mismatch warning**: After a successful POST, the client now compares the server-assigned slug against the requested slug. If they differ (e.g., `category` was requested but `category-0` was created), it emits a warning to surface potential duplicates.
- **Upload throttling**: Adds a 100ms delay between sequential uploads to reduce rate-limit pressure on large batch uploads.

### Test updates

- Added `.times(3)` to nock GET mocks returning 500 status across `changelog/upload.test.ts`, `custompages/upload.test.ts`, and `pages/upload.test.ts` to account for the new retry behavior.
- Increased timeout to 30s on "mix of creates/updates/failures (dry run)" tests where retry backoff across multiple failing slugs exceeds the default 5s vitest timeout.
- All 437 tests pass, 0 failures.

### Production validation

Tested against a real Enterprise project with **585 Markdown files** (the HERE layer-overview documentation set):

| Result | Count |
|---|---|
| Pages created | 578 |
| Pages updated | 6 |
| Pages skipped (no frontmatter) | 1 |
| Failures | 0 |

The initial upload surfaced 5 failures — all caused by typos in the source files' `parent.uri` frontmatter (e.g., `infrasfracture-layers` instead of `infrastructure-layers`), not by the CLI. After correcting the frontmatter, all 5 uploaded successfully.

Notably, the 5 failing files all had `-0`/`-1` suffixed slugs (`category-0`, `poles-0`, `road-surface-markings-0`, `sign-faces-0`, `examples-1`) — artifacts of the exact duplicate-slug problem this PR prevents. These were pages that ReadMe's server had previously created with appended suffixes when a same-slug page already existed.

## Test plan

- [x] `npm run build` passes with no TypeScript errors
- [x] `npx vitest run` — all 437 tests pass
- [x] Dry run against production Enterprise project (585 files) — completes successfully
- [x] Actual upload against production Enterprise project — 584/584 pages created/updated successfully
- [x] Re-upload of previously failing files after frontmatter fix — all 5 succeed


jesseyowell:
@erunion apologies for this! should I just keep these Claude-induced rdme PRs draft going forward? (this is a fairly urgent request in terms of functionality though.. since it's largeish Enterprise groups asking for a fix)

erunion:
@jesseyowell no you're good, i just moved it to a draft so i can pick this up next when i get back to working on rdme stuff.

jesseyowell:
oh just realized you already assigned yourself the Linear ticket involved with this. all good!

erunion:
Closing this out because the API doesn't return 409s.

Issue #1405, opened 2026-01-28
`docs upload` does not correctly propagate "position" parameter


ktham:
## Problem

When uploading guides using `rdme docs upload path/to/guides`, the `position` parameter from the Markdown frontmatter is not correctly being passed to the create/update guide endpoints[^1]

This results in the guide pages being ordered alphabetically rather than the specified order when attempting to sync docs via the `rdme` CLI through the Github Action.[^2]

I'm using the Github Action at this Git revision:
* `readmeio/rdme@5d702394cb1fd1dc3fab6e0769b44b56c9d425d1 # v10.6.0`

## Example

Let's say we have two markdown files:

```markdown
---
title: A Page
slug: a-page
content:
  excerpt: This is a page.
category:
  uri: Get Started
position: 2 # <-- Not being passed correctly to the PATCH guide endpoint
---

## Overview
Foo A
```

```markdown
---
title: B Page
slug: b-page
content:
  excerpt: This is another page.
category:
  uri: Get Started
position: 1 # <-- Not being passed correctly to the PATCH guide endpoint
---

## Overview
Foo B
```

### Expected Result
We should see the pages being ordered like this:
* Get Started
  * B Page
  * A Page

### Actual Result
We actually see the pages being ordered like this:
* Get Started
  * A Page
  * B Page

## Debugging
As a debugging step, I had attempted to manually patch the page's `"position"`, which worked, so there isn't anything wrong with the Patch API, but it looks like the `rdme docs upload` isn't correctly handling this situation.

[^1]: https://docs.readme.com/main/reference/updateguide for `PATCH https://api.readme.com/v2/branches/{branch}/guides/{slug}`
[^2]: https://docs.readme.com/main/docs/rdme

erunion:
This is has been resolved with `rdme@10.7.0`.

Issue #1367, opened 2025-10-02
`position` number in frontmatter of "parent" Markdown file not respected properly


pete-the-pete:
Assume the following file structure and desired `position` of markdown files

1. intro 
2. getting-started 
3. jobs
   1. initiating
   2. updating
   3. closing
4. payments
5. swag
6. contacts

The `payments` page will appear **before** the `jobs` page, even though the position is higher.

erunion:
This has been resolved with v10.7.0.

PR #1449, opened 2026-04-30
fix: `position` sorts should dictate the order that files are uploaded

erunion:
| 🚥 Resolves CX-3309, https://github.com/readmeio/rdme/issues/1405, https://github.com/readmeio/rdme/issues/1367 |
| :------------------- |

## 🧰 Changes

With ReadMe Refactored the `position` value that we use in our API is not _quite_ the same as it was pre-refactored because it dictates the position that your page should be sorted within the `_order.yaml` file that we create for you in Git. If `page-2` has a `position` of `2` and is created _before_ `page-1`, which has a position of `1`, then the `_order.yaml` file will have `page-2` sorted before `page-1` because it came in first.

Pre-Refactored these `position` values were stored as the value you had alongside the page in our database so a file with a `position` of `2` would forever have that position until it explicitly changed.

The work I've done here is to change the way that rdme batches up its API requests to ensure that we make these requests in the order of the `position` value that we have present. If we don't have a `position` then we'll fall back to the `slug` available, sorting it naturally (ensuring that `page-10` is not sorted between `page-1` and `page-2`)

### DIFF
--- a/test/__fixtures__/docs/position-order-siblings/parent-hub.md
+++ b/test/__fixtures__/docs/position-order-siblings/parent-hub.md
@@ -1,0 +1,8 @@
+---
+title: Parent hub
+category:
+  uri: category-slug
+position: 0
+---
+
+Body
--- a/src/lib/syncPagePath.ts
+++ b/src/lib/syncPagePath.ts
@@ -6,7 +6,6 @@ import path from 'node:path';
 
 import chalk from 'chalk';
 import ora from 'ora';
-import toposort from 'toposort';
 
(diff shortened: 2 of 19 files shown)


test/__fixtures__/docs/position-order-siblings/parent-hub.md
```diff
@@ -0,0 +1,8 @@
+---
```
flinehan, line 1: nice new test fixtures

flinehan (approved):
Nice new tests!

PR #1465, opened 2026-05-11
feat: improved messaging around inferred OpenAPI slugs when `--slug` is missing

erunion:
| 🚥 Resolves CX-2743 |
| :------------------- |

## 🧰 Changes

When you upload a new OpenAPI definition with `openapi upload` and don't supply the `--slug` or `--legacy-id` flags we create an inferred slug for the filename based on its path; this results in cases like `~/Desktop/petstore.json` getting created as `desktoppetstore.json`. Now normally this isn't a problem because we don't show these slugs to end-users but if you happen to move `~/Desktop/petstore.json` to `~/code/application/petstore.json` then the next time you invoke `openapi upload` we'll end up creating `codeapplicationpetstore.json` instead of overwriting `desktoppetstore.json`.

What this does is add a new step into the upload process so that  whenever we end up in a situation where we' re going to slugify a nested directory path like `~/Desktop/petstore.json` we confirm with the user that that's the slug they want. It looks something like this:

```
$ ./bin/dev.js openapi upload --key=MY_API_KEY ~/Desktop/petstore.json
Warning: This OpenAPI file is located in a subfolder and no `--slug` was provided.

You are uploading:
  /Users/erunion/Desktop/petstore2.json

In rdme v10, the file path is flattened and used to infer a slug:
  /Users/erunion/Desktop/petstore2.json  →  UserserunionDesktoppetstore2.json

If you intend to update an existing definition, this often creates a new API definition instead of updating an existing one.

We recommend you cancel and rerun with:
  rdme openapi upload /Users/erunion/Desktop/petstore2.json --slug=petstore2.json

To suppress this warning and confirmation in the future, use `--confirm-overwrite`.
✖ Continue anyway? › (y/N)
```

Additionally as part of this I have also added in some improved error handling for when we encounter a case where an API definition fails to upload we'll now tell the user what happened, possibly with some OpenAPI validation errors, instead of a general "Your API definition upload failed with an unexpected error." crash.

### DIFF
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -1,5 +1,7 @@
+import type { SpecType } from '../../lib/prepareOas.js';
 import type {
   APIDefinitionsRepresentation,
+  APIUploadFailureReason,
   APIUploadSingleResponseRepresentation,
   APIUploadStatus,
   StagedAPIUploadResponseRepresentation,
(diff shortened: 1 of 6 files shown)


mosesj-readme (approved):

domharrington (approved):
LGTM when lint is fixed

PR #1499, opened 2026-07-30
fix(openapi): upgrade our bundler to resolver issues with `$ref` resolutions

erunion:
| 🚥 Resolves CX-3747 |
| :------------------- |

## 🧰 Changes

This upgrades our OpenAPI tooling dependencies in order to fix a bug in `@readme/openapi-parser` where OpenAPI bundling could sometimes resolve `$ref` pointer to a pointer that doesn't exist. The issue in question was fixed upstream from us in `@apidevtools/json-schema-ref-parser` in https://github.com/APIDevTools/json-schema-ref-parser/pull/409.


src/commands/openapi/inspect.ts
```diff
@@ -88,14 +88,13 @@ function buildFeaturesReport(analysis: Analysis, features: string[]) {
 function buildFullReport(analysis: Analysis, definitionVersion: string, tableBorder: Record<string, string>) {
   const report: string[] = ['Here are some interesting things we found in your API definition. 🕵️', ''];
 
-  const allowedKeys = ['dereferencedFileSize', 'mediaTypes', 'operationTotal', 'rawFileSize', 'securityTypes'];
-  const sizeKeys = ['rawFileSize', 'dereferencedFileSize'];
```
erunion, line 92: File size statistics were removed from `oas/analyzer` so I've removed them from here as well.

test/__fixtures__/circular-ref-oas/circular-reference-ref-only.resolved.json
erunion, line 1: All of these test schema changes are because circular ref resolution was improved upstream in `@apidevtools/json-schema-ref-parser`.

test/commands/openapi/__snapshots__/inspect.test.ts.snap
```diff
@@ -81,8 +81,15 @@ additionalProperties:
...
+ · #/components/schemas/BodyPart/properties/parent
+ · #/components/schemas/FormDataBodyPart/properties/parent
+ · #/components/schemas/MultiPart/properties/bodyParts/items
  · #/components/schemas/MultiPart/properties/parent
- · #/components/schemas/ZoneOffset/properties/rules"
+ · #/components/schemas/ZoneId/properties/rules
+ · #/components/schemas/ZoneOffset/properties/rules
+ · #/components/schemas/ZoneOffsetTransition/properties/offsetAfter
+ · #/components/schemas/ZoneOffsetTransition/properties/offsetBefore
+ · #/components/schemas/ZoneRules/properties/transitions/items"
```
erunion, line 92: We have improved detection of circular references with these upgrades.

PR #1515, opened 2026-08-12
fix(openapi): treat pending uploads as accepted

HugoHSun:
| 🚥 Resolves [CX-3771](https://linear.app/readme-io/issue/CX-3771/rdme-openapi-upload-times-out-while-the-openapi-definition-continues) |
| :--------------------------------------------------------------------------------------------------------------- |

## 🧰 Changes

`rdme openapi upload` waits for queued API definitions to finish processing. When the polling window ends, it currently reports that the upload timed out and fails CI even if ReadMe confirms that the job is still processing in the background.

CX-3012 previously increased the polling limit from 10 checks (about 3 minutes) to 25 checks (about 10.5 minutes). CX-3771 shows that valid uploads can still take longer. Increasing the limit again would only move the cutoff, make CI wait longer, and keep treating an active backend job as a failed upload.

This changes the exhausted polling-window case to return the last confirmed `pending` or `pending_update` status successfully. The message includes the API definition slug, explains that processing is continuing, warns against retrying while it is still processing, and asks the customer to check ReadMe later for final completion. Actual upload failures and polling request errors remain errors.

## 🧬 QA & Testing

- A newly created definition that remains `pending` after all 25 polls exits successfully, returns the pending status, and includes its slug in the background-processing message.
- An updated definition that remains `pending_update` behaves the same way.
- Definitions that reach `done` keep the existing success behavior, while API and processing failures remain non-zero exits.

### DIFF
--- a/documentation/commands/openapi.md
+++ b/documentation/commands/openapi.md
@@ -220,6 +220,9 @@ DESCRIPTION
   For the best and most explicit results, we recommend using the `--slug` flag to set a slug for your API definition,
   especially if you're managing many API definitions at scale.
 
+  If your API definition is still processing when the polling window ends, the command will exit successfully while
+  processing continues in the background.
+
 EXAMPLES
   You can pass in a file name like so:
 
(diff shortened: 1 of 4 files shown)


HugoHSun:
@coderabbitai review

erunion (approved):

PR #1516, opened 2026-08-12
feat(openapi): add configurable upload timeout

HugoHSun:
| 🚥 Follow-up to [CX-3771](https://linear.app/readme-io/issue/CX-3771/rdme-openapi-upload-times-out-while-the-openapi-definition-continues) |
| :----------------------------------------------------------------------------------------------------------------------------- |

## 🧰 Changes

Adds `--timeout` so customers can choose how long `rdme openapi upload` waits for processing. It defaults to 12 minutes and accepts values from 1 second to 1 hour. The upper limit prevents accidental long-running CI jobs from polling indefinitely.

This is separate from the immediate CX-3771 fix because changing the wait duration is optional configuration rather than part of correcting the false failure.

## 🧬 QA & Testing

- Run an upload that remains pending and confirm the default polling window is 12 minutes.
- Pass `--timeout 3` and confirm the command stops polling after the custom window, then reports that processing continues in the background.
- Confirm `--timeout 0` and `--timeout 3601` are rejected.
- Confirm uploads that finish or fail during polling keep their existing behavior.

### DIFF
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -23,6 +23,7 @@ import prepareOas from '../../lib/prepareOas.js';
 import promptTerminal from '../../lib/promptWrapper.js';
 
 const yamlExtensions = ['.yaml', '.yml'];
+const defaultWaitTimeoutSeconds = 12 * 60;
 
 function isNestedLocalSpecPath(specPath: string): boolean {
   return nodePath.dirname(nodePath.normalize(specPath)) !== '.';
@@ -80,6 +81,14 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
         'If included, use the version specified in the `info.version` field in your OpenAPI definition for your ReadMe project version. This flag is mutually exclusive with `--branch`.',
       exclusive: ['branch'],
     }),
+    timeout: Flags.integer({
+      summary: 'Number of seconds to wait for the upload to finish processing.',
+      description:
+        'Must be between 1 and 3,600 seconds. If the API definition is still processing when the timeout is reached, the command will exit successfully while processing continues in the background.',
+      default: defaultWaitTimeoutSeconds,
+      max: 3600,
+      min: 1,
+    }),
     'working-directory': workingDirectoryFlag,
   };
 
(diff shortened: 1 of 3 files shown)


HugoHSun:
@coderabbitai review

src/commands/openapi/upload.ts
```diff
@@ -80,6 +82,14 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
         'If included, use the version specified in the `info.version` field in your OpenAPI definition for your ReadMe project version. This flag is mutually exclusive with `--branch`.',
       exclusive: ['branch'],
     }),
+    'wait-timeout': Flags.integer({
```
erunion:
```suggestion
    'timeout': Flags.integer({
```

src/commands/openapi/upload.ts
```diff
@@ -115,20 +125,31 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
...
 
     // oxlint-disable no-await-in-loop -- We need to wait between requests to avoid hitting rate limits.
-    while (this.isStatusPending(status) && count < 25) {
-      // oxlint-disable-next-line no-loop-func -- False positive.
+    while (this.isStatusPending(status)) {
+      const remainingMilliseconds = waitTimeoutMilliseconds - elapsedMilliseconds;
+
+      if (remainingMilliseconds <= 0) {
+        break;
+      }
```
erunion, line 140: If we break out of here should we have a special `status` or `reason` for this case rather than inferring on just the default `pending`?

src/commands/openapi/upload.ts
```diff
@@ -115,20 +125,31 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
...
+
+      const pollingDelayMilliseconds = Math.min(1000 * 2 ** count, maxPollingDelayMilliseconds, remainingMilliseconds);
+
       await new Promise(resolve => {
-        // exponential backoff — wait 1s, 2s, 4s, 8s, 16s, 32s, 30s, 30s, 30s, 30s, etc.
-        setTimeout(resolve, Math.min(isTest() ? 1 : 1000 * 2 ** count, 30000));
```
erunion, line 128: I'm having a difficult time following this diff but it looks like we aren't doing this 30s delay anymore? It's now 3600ms?

erunion (changes requested):

src/commands/openapi/upload.ts
HugoHSun: Still 30s, inlined it now to be more clear

src/commands/openapi/upload.ts
HugoHSun: Here the last poll before/at timeout got a pending status, so it would fall to the improved timeout message. I think whether the window was the 12-minute default or user-configured should make no difference here

erunion (approved):

PR #1520, opened 2026-08-21
feat(openapi): surface target project in upload output

jesseyowell:
## 🧰 Changes

`rdme openapi upload` now does a best-effort lookup of the project that the provided API key belongs to (via `GET /projects/me`) and surfaces the project subdomain throughout the upload flow:

- **Overwrite confirmation prompt:** `This will overwrite the existing API definition for petstore.json in the owl-factory project. Are you sure you want to continue?`
- **Success message:** `🚀 Your API definition (petstore.json) was successfully updated in your owl-factory project in ReadMe!`
- **`--dry-run` output:** a `Project: owl-factory` line alongside the existing `Branch`/`Spec Type` lines (`Project: unknown` if the lookup fails)
- **Still-processing message:** `Your API definition (petstore.json) was accepted into your owl-factory project and is still being processed by ReadMe. …`

### Why

An upload with an API key for the *wrong* project previously produced output indistinguishable from a correct upload — the CLI reported success, the user checked their docs site, and nothing had changed. This came up in a support ticket where a customer repeatedly uploaded to a personal test project without realizing their stored key didn't belong to their production project. Naming the target project in the prompt gives users a chance to abort before overwriting anything, and naming it in the success message makes wrong-project uploads immediately diagnosable.

The lookup is failure-tolerant: if `GET /projects/me` errors for any reason, the upload proceeds with the existing message wording (failure is logged in `--debug` output only). It runs in parallel with the existing API definitions fetch, so it adds no latency to the upload flow.

## 🧬 QA & Testing

- New tests cover: project subdomain in the overwrite prompt + success message, project in the still-processing message when the polling window ends, and exact fallback to current wording when the project lookup fails.
- Existing upload tests exercise the fallback path (they don't mock `/projects/me`, so the lookup fails and messages are unchanged) — all snapshots pass untouched.
- Full suite: 518 passed, `tsc` + oxlint/knip/oxfmt clean. (`schemas:check` fails on `next` as well — pre-existing external schema drift, unrelated.)

### DIFF
--- a/src/commands/openapi/upload.ts
+++ b/src/commands/openapi/upload.ts
@@ -214,7 +214,7 @@ export default class OpenAPIUploadCommand extends BaseCommand<typeof OpenAPIUplo
     const [project, existingAPIDefinitions] = await Promise.all([
       // Best-effort lookup of the project this key targets, so output can say where the upload is
       // going (users have uploaded to the wrong project unknowingly). Must never block the upload.
-      this.readmeAPIFetch('/projects/me', { headers })
+      this.readmeAPIFetch('/projects/me', { headers }, { retries: 0 })
         .then(res => this.handleAPIRes<ProjectRepresentation>(res))
         .then(res => res?.data?.subdomain ?? null)
         .catch((e: Error) => {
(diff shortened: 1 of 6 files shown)


jesseyowell:
minor update @erunion, but people can still get confused if they are using the wrong API key, as we've had at least one recorded case of mistaken project upload.

erunion (approved):

erunion (approved):

